PR #152, opened 2016-07-15
Keep database on deploy

tiltec:
Now is the time to preserve the valuable data of our users, instead of scrapping everything on each push.

For this, we need to commit the migrations as well, so don't forget to run `pym makeallmigrations` after model changes.


nicksellen:
Looks good to me :)

PR #157, opened 2016-07-20
Push chat messages to Redis

tiltec:
I put the logic in the serializer. Seemed like the best way to do it. Please comment :)


nicksellen:
I was too late to the party, but no tests?

tiltec:
@nicksellen Yes, I've no idea about testing. I'll try to get into it again.

PR #159, opened 2016-07-21
More testing & resolve warnings

tiltec:
- Remove obsolete files and settings, mostly related to testing
- Resolving deprecation warnings for Django 2.0
- Added first tests for each of the apps (groups, conversations, users)
- Added pytz as requirement


nicksellen:
Looks like a lot of useful changes. I don't think they really belong in one
PR though. Do you explicitly want anything reviewed or discussed? If you
are happy and don't think the change is controversial or needs discussing I
suggest to just commit to master.

tiltec:
Everything should be fine, and given that no one else is developing on the backend right now, I could just commit to master.
However, I did set master as a protected branch, so I *have* to create pull requests. Also to play around with git, GitHub and its features :)

Issue #156, opened 2016-07-20
As a REST-Client I want the 1-n and n-m relations delivered as arrays


Sneezoo:
In these relationships, the right part will give me an ID of the left part, since the left part can have _n_ times of the objects on the right (left _one_ / right _many_):
- User - Group(n:m)
- Group - Store
- Store - PickupDate
- User - Conversation (n:m)

Problem:
The client has to GET **all** objects from the _many_ (Groups, Stores, PickupDates, Conversations) to find the related objects of _one_ object (User, Group, Store).

Example:
If you got a user and would like to know the groups of it, you would have to get all the groups, filter them by member(user)-ids to get them attached to the user.

Solutions:
1. Make the existing GETs return arrays of object-ids if there is a 1:n relationship
2. Create new routes or
3. add filter-parameters to the existing routes to make it possible to filter groups by user-id, stores by group-id, conversations by user-id and pickupDates by store-id.

tiltec:
Does this block you in front-end development right now?
I think you can use the _list all_ endpoints for a long time until we run in performance problems. Then I would choose solution (3) and add the filter parameter.

Sneezoo:
Yeah, I agree there wouldn't be any performance issues by now, but it would look some kind of broken to make the workaround I described.
I just needed the information which option to choose, won't be blocked by this.
Thanks!

tiltec:
The other nice option would be (1). If I understand correctly, we would have the other side of the relation: at GET to a user object returns the IDs of the groups that the user is member in. Is this what you meant?
But then you still would have to GET all the group details.

Sneezoo:
:+1: That's why I put that on top of the list ;-)
That'd be no problem, I can GET the group details specifically for each group. Makes me GETting things straight forward.
I say I would really like to have one of the options implemented, but which one is up to you

tiltec:
Ok, doing some planning about the necessary changes.

# Option (1)

Example:
First some unique IDs for clarification.

- User IDs are 0,1,2,3,4
- Group IDs are 5,6,7,8,9
- Conversation IDs are 10-20
- Group Conversation IDs are 15-19
- Store IDs are 20-29
- Pickup-Date IDs are 30-39

```
user-detail: GET /api/users/(id)/
{ ...,
 groups: [5,7,9],
 conversations: [2: 10, 4: 13], // userid: conversationid
 group-conversations: [15,17]
}

group-detail: GET /api/groups/(id)/
{ ...,
[...]

PR #160, opened 2016-07-21
Remove conversations-by-user endpoint

tiltec:
The same functionality can be achieved by using the the conversations list endpoint and filter by chat participants.

#156 for more details.


tiltec:
(Not sure if it's correct to base this PR on the `more-testing` branch. But if I base it on master right now, there would be too many unrelated changes in this PR)

PR #163, opened 2016-07-22
Store API: Remove pickups field

tiltec:
The pickups can also be accessed through the pickup-dates endpoint, filtering by store ID.
Reduces complexity in the backend. See discussion in #156 .


tiltec:
@Sneezoo @D0nPiano is this acceptable for you?
Basically, you will need to GET /api/pickup-dates/ now and filter by store in the frontend. I will add a API filtering parameter later.

Issue #155, opened 2016-07-20
HTTP request QA


tiltec:
I just found http://pythonhosted.org/HTTPolice/
It is a Linter for HTTP requests and integrates with Django.
Let's see if it's useful.

PR #187, opened 2016-08-10
Check if HTTP requests conform to standards

tiltec:
Issue #155 

Comments welcome! Doesn't seems super useful to me, does it make sense to have it?


nicksellen:
I think we remove a lot of HTTP issues by using drf. Some of the HTTP things change after running through nginx too. Doesn't seem high priority.

PR #195, opened 2016-10-06
Rename deployment to fstool

tiltec:
Also rename database to fstool

### DIFF
--- a/deploy.md
+++ b/deploy.md
@@ -49,7 +49,6 @@ master = true
 processes = 2
 
 socket = /tmp/fstool.sock
-http-socket = /tmp/fstool.http.sock
 
 touch-reload = /tmp/fstool.reload
 
--- a/deploy.md
+++ b/deploy.md
@@ -7,10 +7,10 @@ sudo apt-get install postgresql-server-dev-9.4 python3-dev dev
 Create db user with permissions:
 
 ```
-createuser yunity-dev
+createuser fstool
 sudo -u postgres psql
-> alter user "yunity-dev" with password 'yunity';
-> grant all privileges on database "yunity-dev" to "yunity-dev";
+> alter user "fstool" with password 'fstool';
+> grant all privileges on database "fstool" to "fstool";
 ```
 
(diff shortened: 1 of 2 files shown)


deploy.md
```diff
@@ -48,9 +48,10 @@ module = config.wsgi:application
 master = true
 processes = 2
 
-socket = /tmp/yunity-dev.sock
+socket = /tmp/fstool.sock
+http-socket = /tmp/fstool.http.sock
```
nicksellen: I think I used that so I could run curl commands against it locally on the server to check it responds ok. The non-http socket needs to be requested with uwsgi protocol and command line clients are less/not available.

deploy/remote.sh
```diff
@@ -3,42 +3,43 @@
 set -e
 
 BRANCH=$1
+backend_dir=foodsaving-backend
```
nicksellen: Maybe this is `project_name` and then it can be reused in more places? (ready for the next name change ;))

nicksellen:
How about renaming `fstool` -> `foodsaving`? (then foodsaving.yunity.org...). Otherwise we have two names fstool and foodsaving, and they are essentially the same thing?

deploy.md
tiltec: yeah, found this in bash-history:
`echo -e "GET / HTTP/1.1\r\n" | sudo socat unix-connect:/tmp/yunity-dev.http.sock STDIO`

I guess I can remove it again from the deploy.md

tiltec:
Hopefully we will get the domain foodsaving.org soon (November/December-ish) to host the tool. 

These yunity.org subdomains are just temporarily and for now, it's only a tool and does not contain foodsaving information (how to start your community etc) that I would expect on something called foodsaving.org

So, *no* to rename fstool -> foodsaving right now

deploy/remote.sh
tiltec: I didn't see more places right now, and generally I'm overwhelmed with too generic things ;)

PR #208, opened 2016-10-25
Mail verification and password reset

### LINKED ISSUE #92: Add email verification
Verify users e-mail by sending a mail and requesting the user to click a link.

tiltec:
Integrate support for sending out mails and use it to verify e-mail addresses

Todo:
- [x] make test work (server config missing, no API key)
- [x] write unit & integration test
- [x] setup on fstool.yunity.org (local config file)
- [x] trigger mail verify after mail change
- [x] use email instead of id to request password reset
- [x] change passwords
- [x] check reverification after changing mail address
- [x] include frontend url into mail template for verification
- [x] think about translation support for mail
- [ ] nice mail template
- [ ] decide if it's worth using a serializer for password reset
- [ ] check transaction status of mail
- [ ] log password resets to influxdb
- [ ] add transifex to project
- [ ] decouple the frontend: use a parameter on the `password_reset` request for the email URL prefix
- [ ] consider case: mail is not verified and password reset is triggered

Closes #92 

Sidenote: It may be better to integrate the mail verification tighter into the signup flow.

1. User enters mail to signup
2. we send out one-time token, password is set to "unusable"
3. user clicks on token link to set password

-> Mail is verified, password is set, user is logged in

We can reuse the token for password reset, just without setting the password to "unusable"
The token would be sent to the server together with the new password.

However, the current process is _good enough_, so I'd leave this for a future redesign.


tiltec:
I think this could be merged soon. I will file the remaining tasks as issues :)

PR #226, opened 2016-12-23
Add possibility to protect group with a password

tiltec:
- set password with a PUT request to group (or on group creation)
- send password as POST data on `/join` requests (if required)
- added new fields `password` and `public_description`
- show `description` and `password` field only to group members
- using different serializers depending on membership and operation
- new `protected` field shows password status (set/unset) in list view and retrieve (if not member of the group)


tiltec:
I'm pretty sure that the currently deployed frontend can handle these changes without any regressions. Let's go, merge :)

PR #235, opened 2016-12-30
Add tests, remove unused code, disable click tracking

tiltec:
This is a good pull request.


yunity/userauth/tests/test_api.py
```diff
@@ -32,6 +33,13 @@ def test_wrong_credentials(self):
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(response.data['non_field_errors'], ['Unable to login with provided credentials.', ])
 
+    def test_login_as_disabled_user_fails(self):
+        data = {'email': self.disabled_user.email, 'password': self.disabled_user.display_name}
```
Xbaal: 'password': self.disabled_user.display_name ?
seems wrong ;)

tiltec:
It does, but that's just a simple hack to generate users with a factory and have the password for explicit login. The other way is to use the force_login function, which bypasses the login code.

PR #252, opened 2017-02-23
Added a command to download the latest frontend

nicksellen:
Means you don't have to setup the frontend code locally. Will
download from `https://foodsaving.world/archive.tar.gz` and then
serve it up locally `/static/index.html`.

Needs a cheeky rewrite of `<base href="">` in index.html to match
django settings as it's not possible (or very tricky) to get it
to serve the static files at root.


nicksellen:
It should not have any impact on production deployment (as the frontend files are not actually checked in)

tiltec:
What is the use case of this command? @nicksellen 

A backend developer will most probably use the swagger renderer for interactive tests, as the frontend won't handle API behavior.

nicksellen:
@tiltec use case is for people doing backend-only development but want to see/use the frontend locally.

If changing frontend and backend then they would have both setup anyway.

Would also benefit from `--dev` flag on the command to get the build from `https://dev.foodsaving.world` (not configure in `./deploy.sh` yet though)

PR #258, opened 2017-02-25
style: disable check for empty lines at end of file

tiltec:


tiltec:
It's already in #259 , so I can as well close this..

Issue #269, opened 2017-03-03
Integrate mail templates


tiltec:
Right now, our mails are not very pretty. Let's improve this using text+html mail templates!

Inspired by the format of taiga.io, we already created a set of text-only templates in #268 
Now we need to integrate them. Taiga uses their own [djmail](https://github.com/bameda/djmail), which discovers and loads the templates based on a magic function name (e.g. `mail.passwordreset_request` will use `passwordreset_request*` templates). This seems fine for me.

The important question: does it work together with Anymail, specifically the [status tracking](https://anymail.readthedocs.io/en/latest/sending/tracking/#event-tracking) that we might use in future?

We could also build our own system. Here's a HOWTO on using [anymail with django templates](https://anymail.readthedocs.io/en/latest/tips/django_templates/).

PR #271, opened 2017-03-04
HTML Activation Link Mail Template

anne27:
User-friendly template, contains links to Twitter, Facebook, Google+ and Youtube profiles for Yunity.
Image needs review.
Confirmation link not yet integrated with backend.


tiltec:
First of all, thanks for your great work! This will definitely be useful!

Some remarks:

1. As the template will be used for https://foodsaving.world/, the design could be more similar to the website (e.g. using the logo, brownish colors, Material Design)
2. yunity and foodsaving worldwide are related, but not the same. I think it's unexpected for someone who signs up to see the yunity social links

If that is resolved, we can integrate it with the backend ( #269 ) and the text templates ( #268 ). The directory is ` foodsaving/users/templates/`.

It's very convenient to split the HTML templates into a base part and a content part. An example can be found in the [taiga.io backend code](https://github.com/taigaio/taiga-back/tree/master/taiga/projects/templates/emails). This ensures that all mails have a similar look and feel and reduces duplication.

Feel free to contact me on Slack!

anne27:
Just modified a little to match the formatting in the webapp.

template

tiltec:
I like it very much!

We don't have a privacy statement or terms of service right now, but it would be good to have in future..

I would merge that template later today, what do you think? It still needs splitting into a base template and body (see last comment), but we can do that as soon as the templated mail sending is implemented.

anne27:
@tiltec Sure, you may merge it! I will start working on splitting it soon. :)

PR #281, opened 2017-03-10
Integrate mail templates with code

ShababShahriar:
Uses Jinja2 to render text or html for sending emails.

Connected to Issue #269.


tiltec:
Did you change the templates from Jinja2 to Django Template Language? If yes, do you recommend to use DTL? (and also change the file endings accordingly ;) )

ShababShahriar:
No, the template syntax is still Jinja2, not Django's own. Apparently, they are very much alike (though Jinja2 probably has some advanced features) and the `Django templating system` supports both. Seems good enough to me.

tiltec:
`blocktrans` seems to be DTL-only: [Django docs](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/) vs [Jinja2 docs](http://jinja.pocoo.org/docs/2.9/templates/#i18n-in-templates).

I'm fine with using DTL, there are only a few differences: http://jinja.pocoo.org/docs/2.9/switching/#django
But we should make it clear what we are using :)

foodsaving/users/templates/accountdelete_request-subject.jinja
```diff
@@ -1,3 +1 @@
-{% trans %}
 Request to delete your account
```
tiltec: Is still still translatable? I think you need some translation marker.

You can check by running `manage.py makemessages --ignore=env/* -l en` and see if the generated file `foodsaving/locale/en/LC_MESSAGES/django.po` contains this message.

foodsaving/users/templates/activation-mail.html
```diff
@@ -1,3 +1,5 @@
+{% load i18n %}
+{% blocktrans with display_name=user.display_name %}
```
tiltec: We definitely need to split up this template into a base (containing the skeleton) and a content part (containing the message), otherwise our translators would have to copy and selectively modify all the HTML in Transifex.

foodsaving/users/templates/accountdelete_request-body-text.jinja
```diff
@@ -1,12 +1,13 @@
-{% trans full_name=user.get_full_name(), url=resolve_front_url('delete-account', user.delete_token) %}
+{% load i18n %}
+{% blocktrans with full_name=user.display_name %}
 Hello {{ full_name }},
-we're sad to hear that you want to delete your account on foodsaving.world.
-To finally and irreversibly delete your account click the following link:
+We're sad to hear that you want to delete your account on foodsaving.world.
+To finally and irreversibly delete your account, click the following link:
 
 {{ url }}
```
tiltec: For resolving frontend URLs, I got the inspiration from taiga to collect this at a central place: [definitions](https://github.com/taigaio/taiga-back/blob/master/taiga/front/urls.py) and [helper function](https://github.com/taigaio/taiga-back/blob/master/taiga/front/templatetags/functions.py)

tiltec:
The subject files contain newlines and Django doesn't strip them. Look at the [circleci results](https://circleci.com/gh/yunity/foodsaving-backend/332?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link): `Header values can't contain newlines`

Maybe we strip all newlines from the subject after rendering? Otherwise we will need to be extra careful whenever a new translation comes in.

ShababShahriar:
huh, just came to comment that you were right, and found that you've already mentioned the issue about `blocktrans`. I was reading the DTL documentation and mixed up a bit. Fixing the file extensions

ShababShahriar:
I was trying to see last night what exactly were the issues CircleCi raised but couldn't really find anything. Even the thing about `header containing newlines` can't be seen when I follow the link. Any problem from my end? Do I need an account perhaps?

And yes, we can definitely remove the newline characters from subject. Updating the PR soon.

foodsaving/users/templates/activation-mail.html
ShababShahriar: yeah, makes sense. Not really familiar with Transifex, just googled it.

tiltec:
Maybe you need to login? Should be possible with your github account.

Definitely run your tests locally before you push. `manage.py test` shows all these errors.

foodsaving/users/templates/activation-mail.html
tiltec: Just FYI, here's our project: https://www.transifex.com/yunity-1/foodsaving-tool/dashboard/
It basically helps managing the translation files, we don't need to edit those *.po files manually.

ShababShahriar:
Yeah, that's totally on me. Still didn't set up the database etc. so the tests didn't run fully and I just thought these wouldn't affect any of the tests, especially since they worked well on the shell and weren't really interfering with the existing test cases. Bad decision definitely

ShababShahriar:
```======================================================================
ERROR: setUpClass (foodsaving.management.tests.test_update_pickup_dates.TestUpdatePickupDatesCommand)
  File "/home/ubuntu/virtualenvs/venv-3.4.2/lib/python3.4/site-packages/django/db/backends/base/base.py", line 206, in _cursor
    return self.create_cursor()
  File "/home/ubuntu/virtualenvs/venv-3.4.2/lib/python3.4/site-packages/django/db/backends/postgresql/base.py", line 211, in create_cursor
    cursor = self.connection.cursor()
psycopg2.InterfaceError: connection already closed```

Can't really understand what these are about though. Do they always fail on CircleCi?

File: foodsaving/users/templates/activation-mail.html (hunk 2 above)
Comment (ShababShahriar) [inline foodsaving/users/templates/activation-mail.html]: Nice

**Comment by tiltec** [comment]
These `psycopg2.InterfaceError: connection already closed` are usually follow-up errors on an earlier error. Append `--failfast` to only see the first error.

File: foodsaving/users/templates/accountdelete_request-body-text.jinja (hunk 3 above)
Comment (ShababShahriar) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: `resolve_front_url()` you mean? It couldn't be rendered so I thought we could just send it via `context`. Perhaps we should talk about it a bit? Not sure how the function was working, didn't know where to look for it either.

**Comment by ShababShahriar** [comment]
Wow. So much to learn...

File: foodsaving/users/templates/accountdelete_request-body-text.jinja (hunk 3 above)
Comment (tiltec) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: What do you think about a central place to construct those URLs? `resolve_front_url()` was just the first part of the idea, wasn't implemented yet.

Comment (ShababShahriar) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: Yeah, sounds good definitely. Can we use it directly inside the template though? From what I've seen yesterday, functions can't be processed there.

Comment (tiltec) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: Just checked, isn't possible with DTL. It's possible with Jinja2 though.
If we want to keep DTL, we could pass the constructed URL via context and call `resolve_front_url` in the python file.

Comment (ShababShahriar) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: Hmm that's what came to my mind. Is that okay with you though? Or do you want to switch to Jinja2?

Comment (tiltec) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: Jinja2 seems to give us more options how to pass the data around. If we want to use the taiga way (which works well) then I'd recommend to switch back to Jinja2.

I'm doing a bit more research, just saw that https://github.com/getsentry is a really big Django project.

Comment (tiltec) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: They are passing the URL via context, using Django's `reverse`. We would write our own `resolve_front_url` instead:

https://github.com/getsentry/sentry/blob/master/src/sentry/models/organizationmember.py#L163

Comment (ShababShahriar) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: Okay, reverting to Jinja2. Just to be clear, do we have any other uses for templates other than emails? Definitely not to be used for webpages. Pretty limited scope of application in that sense, no?

Comment (tiltec) [inline foodsaving/users/templates/accountdelete_request-body-text.jinja]: Correct. Not sure if we'll need templates anywhere else in the backend.

File: config/jinja2.py [hunk 4]
```diff
@@ -0,0 +1,18 @@
...
+
+
+class FoodsavingJinja2(Jinja2):
+    app_dirname = 'templates'
+
+
+def environment(**options):
+    env = Environment(extensions=['jinja2.ext.i18n'], **options)
+    env.globals.update({
+        'url': reverse,
```
tiltec: How is that used? Doesn't it get overridden by `url` given in the context?

ShababShahriar: For an urlpattern defined as `url(r'^docs/', get_swagger_view(), name='test-url')`, I believe this allows us to refer it in the template by `{{ url('test') }}`, resulting in an output of `/docs/`.

ShababShahriar:
`django-jinja` seems like a good find. Seems like I'll have to remove pretty much all the environment setup I did so far 😞

config/jinja2.py
tiltec: We will (only?) need to resolve frontend URLs in the templates, right? Similar to the `verify` URL.

ShababShahriar: umm... not sure if I'm understanding you correctly but yes, we don't really need backend urls perhaps. In that case, this seems redundant.

ShababShahriar:
`makemessages` seems to work for only .py files by default. I had to add `-e jinja,py` at the end of the command to account for both .jinja and .py extensions. Didn't add the .po file in the latest commit.

tiltec:
Ah good find! Is there a way to modify the command and add `-e jinja,py` by default?

ShababShahriar:
Had to override the existing `makemessages` command. Wasn't working for an hour or so when I realized that the `django-jinja` package already extended the Command class and was having the final say. 

Wasn't working even after extending the command from `django-jinja` since I put it inside the _django packages_ section in `INSTALLED_APPS` in settings. Since the order is considered by Django to avoid conflicts, I had to move it below the `foodsaving` app. Swapped the whole set of packages for now as you can see in the current `settings.py` file.

Not sure what's up with the codecov failures. Passes all tests as far as I understand.

ShababShahriar:
Test fails due to [this django-jinja issue](https://github.com/niwinz/django-jinja/issues/194). Should work fine as soon as [this PR](https://github.com/niwinz/django-jinja/pull/197) gets merged.

tiltec:
I'll merge this now and it should get deployed to dev.foodsaving.world. We should test this before going in production.

ShababShahriar:
Thanks!

Issue #277, opened 2017-03-06
Save current group for user


tiltec:
The frontend wants to show the users' main group after login. (Currently it chooses any group randomly)

TODO:

- new optional field on user model
- add field to user serializer
- adapt/extend tests

LuisRBarreras:
I am going to work on this issue

PR #286, opened 2017-03-24
Add field current_group on UserModel

LuisRBarreras:
## Create field current_group on UserModel
- Add field current_group foreing_key, null=True
- Create migration
- Update test

This close issue #277


foodsaving/users/models.py
```diff
@@ -65,6 +66,7 @@ class User(AbstractBaseUser, BaseModel, LocationModel):
 
     deleted = BooleanField(default=False)
     deleted_at = DateTimeField(default=None, null=True)
+    current_group = ForeignKey(Group, blank=True, null=True)
```
tiltec: I tend to avoid imports just for foreign key references. You can use string reference there: https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey

nicksellen:
The circleci fails on the flake8 check, you can run it locally to check everything is ok. If you have the virtualenv activated (`source env/bin/activate`) then running `flake8`.

tiltec:
Sorry for merging so late, I was travelling with almost no internet.

This is now deployed on https://dev.foodsaving.world/, next step to make it useful is a frontend implementation. @LuisRBarreras are you also interested in working on that?

PR #299, opened 2017-05-17
Fix276

barryjohnthompson:
This is my fix for issue 276. Let me know what you think.


barryjohnthompson:
This is failing tests because the new 'languages' field does not exist in the database. On my local machine, I fixed this by running 'migrate' and then 'makemigrations'. Do I need to add the resulting migration file to this commit?

tiltec:
Looks really good! Yes, you need to add the migration file for deployment and testing.

The language choices are not mandatory, just for the templates, right?

barryjohnthompson:
Thanks! OK, I'll add that file to this PR when I get the chance.

'choices=LANGUAGES' constrains the valid input for the language field. So, it is not mandatory to specify the language field, but if you do you must choose one of the languages in the LANGUAGES list (you can see the supported languages here: https://github.com/django/django/blob/master/django/conf/global_settings.py). 
I chose to do it this way because Django supports these languages out-of-the-box, so should be more scalable than allowing arbitrary input. I'm happy to revise this though, if needed.

tiltec:
> I chose to do it this way because Django supports these languages out-of-the-box, so should be more scalable than allowing arbitrary input.

I would really just drop this check. There are many django translations, but these are in no way related to our frontend translations (these are only in the frontend repo).

barryjohnthompson:
I've updated the commit and removed the languages check.

tiltec:
Great! Though the migrations file still contains all the language choices, can you delete it and add ths updated one?

barryjohnthompson:
My bad! I think that should do it now.

tiltec:
Done! Next step is implementing the language persistence logic in the frontend, are you also interested in this?

PR #303, opened 2017-05-20
Enable html template

### LINKED ISSUE #292: Enable HTML mails (with plaintext alternative)
UPDATE:

We currently only send out text mails. There is one HTML template around, we should send it and create more. Be sure to also send plaintext alternative for non-HTML clients.

- [ ] add HTML content (see the code from @ShababShahriar below)
- [ ] tests

OLD CONTENT:

After signup or mail change: the mail "Verify your mail address" does contain only raw HTML code:

```
<html>
  <head>
    <title>WELCOME</title>
...
```
Instead, it should contain both the simple plain-text message (for clients who block HTML) and nicely rendered HTML.

(Also, if I only change my mail address or request the verification code again, I don't expect another "welcome" mail)

tiltec:
Closes #292 

But the HTML code has bad style and is hard to maintain, I'm unwilling to merge it in this state.


tiltec:
I think the way forward is to have very simple HTML templates that can be prettified and extended later.

nicksellen:
I was just looking for what are the cool ways to do email templating in 2017 [1].

I think https://mjml.io/ looked really good [2]. You write templates with custom tags, then use their nodejs-based tool to turn them into html. That wouldn't include any templates params, as they would be pre-compiled to html.

So, initially I'd probably propose to not including all the tooling, just use their online editor [3] to create ONE simple template that we can use for all emails, and include a few variables, then use normal django templating.

[1] https://colorlib.com/wp/responsive-html-email-templates/
[2] https://mjml.io/documentation/#basic-layout
[3] https://mjml.io/try-it-live/templates/basic

tiltec:
Should we issue this as a starter task? Should be done in the frontend repo.

tiltec:
Closing because of inactivity, will leave the branch around to pick up work again.

Issue #232, opened 2016-12-25
Check if default permission_classes is always defined and returned


tiltec:
Otherwise list_route and detail_route permission classes are not active

yogesh-kamble:
Do we need to add default permissions classes in settings or for each route we have to add permission_classes

tiltec:
Sorry, this issue needs more explanation. This is only relevant if `permission_classes` is used with list/detail_route decorators. Then `get_permissions()` needs to return the default permission class. I *think* this is the case in our codebase, but I didn't verify yet. This is what the issue is about.

yogesh-kamble:
> @list_route(
        methods=['POST'],
        permission_classes=(IsAuthenticated,)
    )
    def resend_verification(self, request, pk=None):
        "resend verification mail"
        if request.user.mail_verified:
            return Response(status=status.HTTP_400_BAD_REQUEST,
                            data={'error': 'Already verified'})
        request.user.send_new_verification_code()
        return Response(status=status.HTTP_204_NO_CONTENT, data={})

so for above route get_permissions should return the default permissions which is defined in settings.py as well as "IsAuthenticated" which is defined in here ?

tiltec:
By `default`, I mean `self.permission_classes`. Have a look at any `get_permission_classes` method in our codebase. I don't take the value from settings.py into account.

yogesh-kamble:
created this pull request https://github.com/yunity/foodsaving-backend/pull/306 
please review it.. if changes are correct then will make same changes in other get_permission_classes method.

tiltec:
Seems no change is necessary.

PR #306, opened 2017-06-14
Issue 232

yogesh-kamble:


tiltec:
I think I need my laptop to look a bit more into it. I think this code was just fine before,  now the tests fail. Do you know why?

yogesh-kamble:
fixed the unit test failure issue

foodsaving/users/api.py
```diff
@@ -42,6 +42,9 @@ def get_permissions(self):
             self.permission_classes = (AllowAny,)
         elif self.action in ('list', 'retrieve'):
             self.permission_classes = (IsAuthenticated,)
+        else:
+            # Issue 232 default permission classes should always be returned
+            self.permission_classes += self.__class__.permission_classes
```
tiltec: I don't understand what this line doing. Sorry for not being helpful, I'm still travelling and can just occasionally check my phone.
However, I think this code doesn't need to be changed, or what do you want to achieve with it?

yogesh-kamble: right now in case of /api/users/verify_mail/ self.permission_classes return 
`(<class 'foodsaving.users.permissions.IsNotVerified'>, <class 'rest_framework.permissions.IsAuthenticated'>)` 
this tuple is not returning default permission class which is defined at Class level.
and with the above code ... self.permission_classes return 

`(<class 'foodsaving.users.permissions.IsNotVerified'>, <class 'rest_framework.permissions.IsAuthenticated'>, <class 'foodsaving.users.permissions.IsSameUser'>)`

here IsSameUser is permission classes which is defined at class level and I am assuming the classes which is defined at class level is default permission and should always be returned.

tiltec: Verify_mail doesn't have get a user ID, therefore issameuser would throw an exception. I also understand better how permission classes are evaluated: from first to last, and only when necessary.
We have a misunderstanding about the default, it just means that every get_permissions() method should return self.permission_classes by default, to allow for the definition of other classes in the decorator. In this regard, the code above was just fine, but I didn't check the other APIViews yet. this is what the issue was about.
Maybe I should mark such codestyle-only issues, similar to as in the frontend repo.

tiltec:
See #232

PR #331, opened 2017-08-01
Added Dockerfile and changed README

mddemarie:


tiltec:
Great, thanks!
Can you move the instructions from the Dockerfile into README? Ideally new devs don't need to read the dockerfile.

And yes, it would be great if we could remove the "non-docker" setup instructions, but for now I'd leave them.

id-gue:
Hi Tilmann, I moved the instructions. Is it okay like this?

tiltec:
Seems good!

PR #334, opened 2017-08-06
Add pytest and friends

nicksellen:
Ok! Should work like this:

Install dev dependencies:
```
pip-sync *.txt
```

Run all tests as normal (ish):
```
pytest
```

Run all tests from a specific folder:
```
pytest foodsaving/groups/
```

Watch tests and only rerun changed things:
```
ptw -- --testmon
```

Watch tests in a specific folder and only rerun changed things:
```
ptw foodsaving/groups/ -- --testmon
```

You can still use:
```
./manage.py test
```

I also added a bunch of stuff into `manage.py` which speeds up test execution.


nicksellen:
However, if it's just speed you're after, the fastest test execution I've managed so far is __4.7 seconds__ with:

```
(env) ada/nick: time ./manage.py test --parallel 4 --keepdb
Using existing test database for alias 'default'...
Using existing clone for alias 'default'...
Using existing clone for alias 'default'...
Using existing clone for alias 'default'...
Using existing clone for alias 'default'...
System check identified no issues (1 silenced).
.................................................................................................................................................................................................................................................
Ran 241 tests in 3.431s

OK
Preserving test database for alias 'default'...
Preserving test database for alias 'default'...
Preserving test database for alias 'default'...
Preserving test database for alias 'default'...
Preserving test database for alias 'default'...

real	0m4.772s
user	0m9.497s
sys	0m0.620s
```

nicksellen:
So, a simpler method and it actually seems to run about as fast as the one that only runs changes..

```
find foodsaving -name '*.py' | entr ./manage.py test --parallel 4 --keepdb
```

From https://stackoverflow.com/a/38521930

tiltec:
The `entr` tool seems to be a good choice, can it be added to the dockerfile? Ideally we would wrap it in a management command.

On the other side, I guess pytest has nicer reporting.

nicksellen:
I think I'll just use `entr` - I also started using `--fastfail` (or is it `--failfast`) which gives a really quick feedback loop for failing tests (~2-3 seconds in most cases).

So perhaps we don't need all this (`tblib` is useful though, also I made the dev deps have their own .in file...).

tiltec:
I'd remove pytest and add a hint about `entry` and `--parallel` to the README. There's already a section on how to speed up test.

Furthermore, the test_settings can be refactored similar to #333 

I thought about adding a convenience command `manage.py test_watch`, but it seems easier to set a bash alias.

On a side note, I just noticed that some tests fail if I run in reverse (`-r`). Might be my fault for having unwanted side effects in the tests.
- test_create_fails_if_default_language_is_not_set
- test_create_fails_if_email_is_not_unique 
- test_swagger_coreapi (this seems to be a follow-up error since the DB connection shut down)

nicksellen:
Yeah I think we can close this. And add a clean PR for test settings
(removing it from conversations too)

On 9 Aug 2017 13:05, "Tilmann Becker" <notifications@github.com> wrote:

I'd remove pytest and add a hint about entry and --parallel to the README.
There's already a section on how to speed up test.

Furthermore, the test_settings can be refactored similar to #333
<https://github.com/yunity/foodsaving-backend/pull/333>

I thought about adding a convenience command manage.py test_watch, but it
seems easier to set a bash alias.

On a side note, I just noticed that some tests fail if I run in reverse (-r).
Might be my fault for having unwanted side effects in the tests.

   - test_create_fails_if_default_language_is_not_set
   - test_create_fails_if_email_is_not_unique
   - test_swagger_coreapi (this seems to be a follow-up error since the DB
   connection shut down)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/yunity/foodsaving-backend/pull/334#issuecomment-321224021>,
or mute the thread

PR #337, opened 2017-08-08
WIP: User Feedback for PickUps

id-gue:


foodsaving/stores/models.py
```diff
@@ -45,7 +57,7 @@ class PickupDateSeries(BaseModel):
         related_name='series',
         on_delete=models.CASCADE
     )
-    max_collectors = models.PositiveIntegerField(blank=True, null=True)
+    max_collectors = models.Posit iveIntegerField(blank=True, null=True)
```
dmt: typo?

foodsaving/stores/models.py
```diff
@@ -30,6 +30,18 @@ def __str__(self):
         return '{} ({})'.format(self.name, self.group)
 
 
+class Feedback(BaseModel):
+    class Meta:
+        unique_together = ('weight', 'comment')
```
tiltec: You can delete `unique_together` (and the whole `class Meta`) as it would have unwanted effects. Remember our Slack chat about it?

foodsaving/stores/models.py
tiltec: Causes the test failure: https://circleci.com/gh/yunity/foodsaving-backend/521

id-gue: it's already fixed but I created a new problem so I didn't want to upload the changes...

foodsaving/stores/tests/test_feedback_api.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.now = timezone.now()
+        cls.member = UserFactory()
+        cls.group = GroupFactory(members=[cls.member, ])
+        cls.store = StoreFactory(group=cls.group)
+        cls.pickup = PickupDateFactory(store=cls.store)
+        cls.series = PickupDateSeriesFactory(max_collectors=3, store=cls.store)
```
tiltec: You won't need a series to test pickup feedback. I always like to remove code when it's not needed :) By removing cls.series, you can also get rid of cls.series.update_pickup_dates and cls.now

foodsaving/stores/tests/test_feedback_api.py
```diff
@@ -0,0 +1,48 @@
...
+        cls.series.update_pickup_dates(start=lambda: cls.now)
+
+    def test_new(self):
+        """
+        User is trying to give a feedback for a pickup
+        that he didn't do (= not part of the collectors)?
+        """
+        user_feedback = Feedback()
```
tiltec: What did you try to do here?
Maybe you could formulate in words first what the test should do, e.g. something like this:

```py
"""
setup:
1. create user
2. create group with the user as member
3. create a store within that group
4. create a pickup within that store that is in the past (to make allow giving feedback)
5. add user to the pickup as collector

test:
1. log user in
2. user gives feedback to pickup
3. make sure that response is valid
4. make sure that feedback has been created
"""
```

This way it's easier for you to have an overview of the tasks that the test needs to do and you don't need to write it directly into code. It's also easier to spot why a test fails: either there were wrong assumptions what the tests needs to to, or it was just not the correct code to do it.

By the way, the steps I described above are for the _correct_ case, which I would always write before the edge cases - it's simply more fulfilling to me to see the code doing the right thing :)

foodsaving/stores/serializers.py
```diff
@@ -229,3 +232,30 @@ def validate_weeks_in_advance(self, w):
...
+        model = FeedbackModel
+        fields = ['id', 'given_by', 'about', 'weight', 'comment']
+        # update_fields = ['about', 'weight', 'comment']
+
+    def create(self, validated_data):
+        # validated_data['given_by'] = self.context['request'].user
+        feedback = super().create(validated_data)
+        return feedback
+
+        # for creating a history event for adding feedback
```
tiltec: Yes, you can just delete it. Less distraction!

foodsaving/stores/serializers.py
```diff
@@ -229,3 +232,30 @@ def validate_weeks_in_advance(self, w):
...
+
+
+class FeedbackSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = FeedbackModel
+        fields = ['id', 'given_by', 'about', 'weight', 'comment']
+        # update_fields = ['about', 'weight', 'comment']
+
+    def create(self, validated_data):
+        # validated_data['given_by'] = self.context['request'].user
```
tiltec: This should be the correct code. You probably need to configure the serializer to not expect `given_by` as input, by setting it as read_only.

It should work similar to here: https://github.com/yunity/foodsaving-backend/blob/95bee2e924dac544e84f6820c4071d6c2d2d2214/foodsaving/invitations/serializers.py#L14

PR #340, opened 2017-08-09
Add server config generator

nicksellen:
Ok, I made it a generator instead of just fixed files.

Might be a good idea to move some of the server config over to ansible, it's great for this kind of stuff... and get do it incrementally.

Websockets seem to work through all the layers too yay! You can try it in the console with (and then just see that it had a 101 protocol upgrade thing - can't actually check messages actually pass properly yet...):

```js
var url = [
  window.location.protocol.replace(/^http/, "ws"),
  "//",
  window.location.host,
  "/api/ws"
].join("");
new WebSocket(url)
```


nicksellen:
This bit is nice:

> One of the benefits of decoupling the client connection handling from work processing is that it means you can run new code without dropping client connections; this is especially useful for WebSockets.
http://channels.readthedocs.io/en/stable/deploying.html

So it should mean that during restarts, if a client sent a websocket message, but the workers were already shutting down, it will just hang around in the redis queue, then when the new worker is loaded, it can read the message and handle it. The client will stay connected the whole time and never know something was restarted.

tiltec:
Daphne handles the client connections, right? So there seem two kinds of restarts:

1. After changing "our" code base
2. After daphne upgrade

What needs to be done in case (1)? So far the [deploy script](https://github.com/yunity/foodsaving-backend/blob/master/deploy/remote.sh) replaces the code and then does uwsgi _touch-reload_. With the workers, would the deploy script need to restart the individual workers?

```
systemctl restart foodsaving-world-worker@1
systemctl restart foodsaving-world-worker@2
systemctl restart foodsaving-world-worker@3
...
```

or can it simply do:

```
systemctl restart foodsaving-world-worker
```

Another point: how can we do that without root privileges? There's a way to allow `sudo` only for certain commands: https://unix.stackexchange.com/questions/192706/how-could-we-allow-non-root-users-to-control-a-system-d-service

After a daphne upgrade (2), we would need to restart `foodsaving-world-daphne`.
I guess we can do that manually for now. (It would also drop all client connections). It would make sense to me to pin the daphne version inside the `requirements.in` file and upgrade separately from other packages.

nicksellen:
> With the workers, would the deploy script need to restart the individual workers?

Yes, we have to handle the reloading ourselves. uwsgi was managing that before by stopping/starting processes. The [runworker command](https://github.com/andreif/django-channels/blob/master/channels/management/commands/runworker.py) is very simple.

Because of the queueing it will not take the website down and all requests will be serviced.

If they receive a SIGTERM they'll shutdown cleanly [1], which is the systemd default [2].

> systemctl restart foodsaving-world-worker

I/we can look to see if we can define a master service that allows all the seperate instance service to be controlled as one...

> After a daphne upgrade (2), we would need to restart foodsaving-world-daphne

I would not worry too much about automating this. It would also need to be restarted if we changed the channel layer config, which won't happen very frequently. I doubt minor version updates would make much difference either. So, I would keep to doing it manually for now.

[1] http://channels.readthedocs.io/en/stable/deploying.html#deploying-new-versions-of-code
[2] https://www.freedesktop.org/software/systemd/man/systemd.service.html

nicksellen:
> I/we can look to see if we can define a master service that allows all the seperate instance service to be controlled as one...

`BindsTo` seems the directive to use [1]. I'll try it out on the foodsaving-world-channels config

[1] https://www.freedesktop.org/software/systemd/man/systemd.unit.html#BindsTo=

nicksellen:
or `PartOf` ... I'll play with them

nicksellen:
Ok, I added a target that can control all the worker instances. The `PartOf` makes the service instance start/stop/restart when the target is.

I also removed the link between daphne and the workers, I think it keeps it simpler as they will be started/stopped for very different reasons and don't want to be knee deep in the systemd docs all the time trying to understand the interactions.

server-configs/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Put the systemd files in `/etc/systemd/system/`.
+
+And the nginx one in `/etc/nginx/sites-available/`.
+
+## Useful commands on server
+
+Start the service:
+```
+systemctl start foodsaving-world-channels-daphne.server
```
tiltec: This should end on `.service` instead of `.server`, right?

deploy/deploy-systemd.sh
```diff
@@ -0,0 +1,64 @@
...
+  name=foodsaving-world
+else
+  name=foodsaving-world-dev
+fi
+
+function restart-workers() {
+  sudo systemctl restart "${name}-worker.target"
+}
+
+backend_dir="/var/www/${name}/www"
```
tiltec: Do you intend to change the backend dir from `/home/deploy/` to `/var/www/`? If yes, why?

You would need to update the cronjobs as well: 

image

nicksellen: Yes, I did intend it. To make it consistent with all the other apps on the server. Good point for the cronjobs. Although perhaps we could go all-out on the systemd, using [timers](https://wiki.archlinux.org/index.php/Systemd/Timers#As_a_cron_replacement).

server-configs/README.md
nicksellen: Ah yeah, typo.

deploy/deploy-systemd.sh
nicksellen: Or, perhaps add celery finally, and setup the timer things in there.

nicksellen: Either way, it's nice to have these things defined inside the app, or at least the repo if possible.

tiltec: Hm yes. Well if we go for celery in the foreseeable future, it would need another systemd service. Or this recommended python alternative: http://supervisord.org
Which could also make sense to use for daphne and the workers.

nicksellen: Personally I'm not into these language-ecosystem-specific developer-friendly kind of process managers (god, pm2, supervisord, etc) - the os already does all this stuff, systemd, upstart, smf, etc..

tiltec: I remember looking into supervisord some time ago and it looked quite good. But right now I don't see any reason to use it over systemd.
Config has yet another format with similar complexity than systemd service files: https://github.com/celery/celery/blob/master/extra/supervisord/supervisord.conf

tiltec: Celery will become necessary to send out mails and notifications triggered by user interaction. But maybe not everything at once.

nicksellen: Celery looks easy to add in the future, but unrelated to this change now.

nicksellen:
So, proposal: 
1. edit deploy.sh to use the systemd deployment method for dev branch only
2. switch over nginx config to use channels backend
3. switch over cronjobs to use new path
4. merge to master

tiltec:
seems good to me!

PR #342, opened 2017-08-10
WIP: User Feedback for PickUps Tests

mddemarie:
We do a new WIP Pull request because we had to remove our GitHub branch feedback3 due to local merge conflict. But we saw the suggestions and we integrated them into our code.

Tiltec: we added 2 tests in stores/tests/test_feedback_api and they both work. There are 2 things we did not finish today: 1. pull last changes from master (we have a merge conflict there) 2. integrate read_only for given_by in serializers.py.


tiltec:
I can't wait for the next test which is hopefully `test_create_feedback_as_user` and it shows how a feedback gets created 🍾 🎉  :)

foodsaving/stores/serializers.py
```diff
@@ -229,3 +232,40 @@ def validate_weeks_in_advance(self, w):
...
+
+class FeedbackSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = FeedbackModel
+        fields = ['id', 'weight', 'comment', 'about', 'given_by']
+        # read_only_fields = ('given_by',)
+
+        """
+        Tilmann suggested to make 'given_by' read_only - but the following code produces errors
+        extra_kwargs = {
```
tiltec: I would be interested to see what errors you get. Probably the database complains that the `given_by` field is missing a value, as it is non-nullable.

To solve this issue, you need to provide then `given_by` field when creating the entry. Override this `create()` method for that.

```
        validated_data['given_by'] = self.context['request'].user
        return super().create(validated_data)
```

I just noticed that you discovered the new `read_only_fields` [shortcut](http://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields). Great, please use that instead of `extra_kwargs`!

foodsaving/stores/tests/test_feedback_api.py
```diff
@@ -0,0 +1,133 @@
...
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN, response.data)
+
+    def test_list_feedback_fails_as_non_group_member(self):
+        """
+        Non-Member is NOT allowed to see list of feedback
+        """
+        self.client.force_login(user=self.user)
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND, response.data)
```
tiltec: I would expect an empty list instead of `404`, similar to [here](https://github.com/yunity/foodsaving-backend/blob/3280d295173176a57ea0a5049bc1ad8686143ed7/foodsaving/stores/tests/test_pickupdates_api.py#L437)

Explanation: the request goes to `/api/feedback/`, which is not group-specific. It will return the list of all feedback the user has access to, which is an empty list in this case.

foodsaving/stores/api.py
```diff
@@ -59,6 +69,46 @@ def perform_destroy(self, store):
...
+    # queryset = PickupDateModel.objects.filter(deleted=False)
+    permission_classes = (IsAuthenticated,)
+
+    # tried with def detail but returned 1 test failure
+    # @detail_route(
+    #    methods=['GET'],
+    #    serializer_class=FeedbackSerializer
+    # )
+
+    # returns 1 failure in tests
```
mddemarie: Thank you, this is definately helpful! 👍

foodsaving/stores/tests/test_feedback_api.py
mddemarie: We have done that yesterday evening. You can look it up in our follow-up commit that I pushed.

foodsaving/stores/serializers.py
mddemarie: Yes, we are getting errors also for `read_only_fields`. We will try it then with `create()` to avoid errors.

foodsaving/stores/models.py
```diff
@@ -30,6 +30,13 @@ def __str__(self):
         return '{} ({})'.format(self.name, self.group)
 
 
+class Feedback(BaseModel):
+    given_by = models.ForeignKey('users.User', on_delete=models.CASCADE, related_name='user_feedback')
+    about = models.ForeignKey('PickupDate')
```
nicksellen: I find the name `about` unintuitive, and it sounds like it's maybe a generic relation (but it's not) - how about `pickup`?

foodsaving/stores/serializers.py
```diff
@@ -229,3 +232,41 @@ def validate_weeks_in_advance(self, w):
...
+            raise serializers.ValidationError(_('You are not member of the store\'s group.'))
+        return user_id
+                """
+    # Is the user a member of the group?
+    # def validate_given_by(self, user_id):
+    #     if user_id is in groups
+
+    #     if not self.context['request'].user.groups.filter(id=group_id).exists():
+    #         raise serializers.ValidationError(_('You are not member of the store\'s group.'))
+    #     return user_id
```
nicksellen: Remove commented out code

foodsaving/stores/api.py
```diff
@@ -59,6 +67,51 @@ def perform_destroy(self, store):
...
+
+    """
+
+    # Looked like correct one for retrieving data:
+    # def get_permissions(self):
+
+    #    if self.action == 'retrieve':
+    #        self.permission_classes = (IsAuthenticated, IsMember,)
+
+    #    return super().get_permissions()
```
nicksellen: Remove commented out code

foodsaving/stores/models.py
tiltec: I'm ok with both, change it if you agree with Nick and you feel like doing refactoring.

mddemarie: we thought the same. `about` is too unintuitive. we will change it then.

foodsaving/stores/serializers.py
```diff
@@ -229,3 +232,41 @@ def validate_weeks_in_advance(self, w):
...
+
+    def validate_about(self, about):
+        user = self.context['request'].user
+        pickup_date = about  # PickupDateModel.objects.get(pk=about)
+        group = pickup_date.store.group
+        # is the member assiged to the pickup?
+        if user not in group.members.all():
+            raise serializers.ValidationError(_('You are not member of the store\'s group.'))
+        # if user is in self.context['request'].
+        if about not in self.context['request'].user.pickup_dates.all():
```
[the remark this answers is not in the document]
mddemarie: Ok, we know now what is needed to refactor. Thank you.

foodsaving/stores/models.py
mddemarie: And are you sure that the name `pickup` will not crash with `pickup` at different places in code?

PR #343, opened 2017-08-10
run circleci tests with two processes

tiltec:


tiltec:
Oh well this breaks codecov. There are ways around it, but it doesn't seem worth the effort: https://coverage.readthedocs.io/en/coverage-4.2/subprocess.html

PR #348, opened 2017-08-14
added commands for run server, test data etc. in dockerfile and readme

mddemarie:
The dockerfile simplies to run a docker from bash and it is no more necessary to have a docker always open in terminal while doing other things. New commands are now in Dockerfile for test data and runserver.


nicksellen:
I suggest you name the container when you create it, then you can include that name in the logs command. Its nice when you can just copy and paste the readme instructions exactly.

tiltec:
Oh Nick, that was coincidence. Didn't mean to shut you down.

mddemarie:
@nicksellen I would like to do that but the thing is that you get that name randomly and I cannot know now what name it will be like

mddemarie:
@tiltec yep, we did some more changes today with Daniel and thus I deepen my knowledge about Docker :)

tiltec:
Not being Nick, but docker run has a `--name` option: 
https://docs.docker.com/engine/reference/run/#container-identification

`docker run --name my-backend backend`

PR #356, opened 2017-08-30
Add push subscription infrastructure

nicksellen:
This implements https://github.com/yunity/foodsaving-backend/issues/350

* endpoints under `/api/subscriptions/push/`
* supports create, get, list, delete
* prevents duplicate subscriptions (unique on user+token)
* on each conversation message send the message content to everyone but the author
* new notifications for a given conversation will replace previous notifications on their phone (using `tag` feature)
* will delete invalid subscriptions (if fcm returns `InvalidRegistration` message for that token)

Things it does not do that I am not so concerned to implement for now:
* only send notifications if the user is not actively using the chat via websocket


foodsaving/subscriptions/serializers.py
```diff
@@ -0,0 +1,33 @@
...
+            'platform'
+        ]
+        validators = [
+            UniqueTogetherValidator(
+                queryset=PushSubscription.objects.all(),
+                fields=('user', 'token')
+            )
+        ]
+
+    user = UserIdSerializer(
```
nicksellen: Adding this was the only way I could get it to respect the `UniqueTogetherValidator`. I don't actually want the user info in the serializer, and adding/using `UserIdSerializer` was just to minimize the extra stupid data.

I suspect there is a better way to solve it.

foodsaving/subscriptions/serializers.py
```diff
@@ -0,0 +1,33 @@
...
+        read_only=True,
+        default=CurrentUserDefault()
```
nicksellen: This `CurrentUserDefault` is nice, we can replace some custom things we wrote to do that I think.

foodsaving/subscriptions/tests/test_receivers.py
```diff
@@ -56,3 +62,34 @@ def tests_receive_message_on_leave(self):
...
+        conversation.join(user)
+        conversation.join(author)
+
+        # add a push subscriber
+        PushSubscription.objects.create(user=user, token=token, platform=PushSubscriptionPlatform.ANDROID)
+
+        # add a message to the conversation
+        ConversationMessage.objects.create(conversation=conversation, content=content, author=author)
+
+        # we can't check it was received but the check_json_data above will at least check it sent the write info
```
nicksellen: typo: write -> right

foodsaving/subscriptions/serializers.py
tiltec: Seems a good workaround, maybe add a short comment in the code?

PR #357, opened 2017-08-30
Add user group through model

### LINKED ISSUE #354: Add user/group through model
So https://github.com/yunity/foodsaving-backend/issues/353 can be done.

I think django does not support this change, so might have to hack around a bit.

nicksellen:
Fixes #354

Seems to work! I tried it on freshly migrated databases (via running the tests), and via databases which have the existing schema and contain data by doing:

```
git checkout master && \
  ./manage.py reset_db --noinput && \
  ./manage.py migrate && \
  ./manage.py create_sample_data && \
  git checkout add-user-group-through-model && \
  ./manage.py migrate
```

... and it applies the migrations ok and all the data stays there. One field is added (created_at field on the membership table).

All the tests work, but I have barely tried the interface (just created an account in swagger), so it's a good trial of how good our test coverage really is ;)

It would be useful if somebody would try it out though.

And before deployment make sure there is a backup.


foodsaving/groups/factories.py
```diff
@@ -14,7 +14,7 @@ def members(self, created, members, **kwargs):
             return
         if members:
             for member in members:
-                self.members.add(member)
+                GroupMembership.objects.create(group=self, user=member)
```
tiltec: This occurs quite a lot. Maybe use the `add_member(user)` helper method to the Group model and modify it to send out the signal only if a parameter is set?

foodsaving/groups/migrations/0002_group_members.py
```diff
@@ -16,9 +17,32 @@ class Migration(migrations.Migration):
...
+            ],
+            options={
+                'db_table': 'groups_group_members',
+            },
+        ),
         migrations.AddField(
             model_name='group',
             name='members',
-            field=models.ManyToManyField(to=settings.AUTH_USER_MODEL),
+            field=models.ManyToManyField(related_name='groups', through='groups.GroupMembership', to=settings.AUTH_USER_MODEL),
```
tiltec: This change pretends that there was never a version _without_ the `through` model. Makes sense to me!

nicksellen: Shhhhh don't tell anyone. It's always been this way... ;)

foodsaving/groups/factories.py
nicksellen: I thought about it, but didn't want to open up the topic at this point.

At the moment our signal use is inconsistent as I've used all the built in django ones, but the existing code had defined custom ones. I didn't think about it much lately.

https://stackoverflow.com/a/26490827 has a way to disable signals during factory setup.

Maybe using the add_member helper and suppressing the signals would be better at the moment.

nicksellen: I pulled it out into it's own issue, because the solution was not clear to me https://github.com/yunity/foodsaving-backend/issues/361

So, if you're happy I would propose to merge this as it is now, and rework it depending on the outcome of #361?

PR #359, opened 2017-08-30
Remove random bit of code

nicksellen:
I have no idea what that is doing there. It is not called from what I can tell, and the tests work without it. And it makes no sense whatsoever, create a message each time you retrieve the conversation info??!?!?!


tiltec:
Good catch! Let's aim for 100% coverage :)

Issue #353, opened 2017-08-29
Add rules/permissions system


nicksellen:
To implement https://github.com/yunity/foodsaving-frontend/issues/550 we need a way of defining role permissions, e.g. _is this user an admin of the group?_

Django includes object-level permissions concept:

> Django’s permission framework has a foundation for object permissions, though there is no implementation for it in the core. That means that checking for object permissions will always return False or an empty list (depending on the check performed). An authentication backend will receive the keyword parameters obj and user_obj for each object related authorization method and can return the object level permission as appropriate.
https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#handling-object-permissions

Django Rest Framework supports using Django object-level permissions: http://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions

The _standard_ implementation is https://github.com/django-guardian/django-guardian - this adds some database models to store the permissions, and can be used like (from the readme example):

```py
>>> from django.contrib.auth.models import User, Group
>>> jack = User.objects.create_user('jack', 'jack@example.com', 'topsecretagentjack')
>>> admins = Group.objects.create(name='admins')
>>> jack.has_perm('change_group', admins)
False
>>> from guardian.models import UserObjectPermission
>>> UserObjectPermission.objects.assign_perm('change_group', jack, obj=admins)
<UserObjectPermission: admins | jack | change_group>
>>> jack.has_perm('change_group', admins)
True
```

An _alternative_ implementation is https://github.com/dfunckt/django-rules - this uses rules/predicates to check permissions - which may in turn use your existing database models to check for permissions gained due to the model relationships (e.g. can edit store? = `user.groups.includes(store.group)`) without needing to add additional entries to the permissions table as in `django-guardian`.

I like the django-rules philosophy.

However, we don't actually _need_ anything extra as Django Rest Framework already sets us implement arbitrary logic in the permission classes, so why add anything else? A few reasons:
* makes the rules/permissions easily reusable across REST API viewsets 
* separates the permissions from the API - we maybe want to check permissions elsewhere in the code (in websocket handlers, in workers, in cron tasks, in management commands, in serializers)
* buys into the django object-level permissions ecosystem, perhaps there are other tools that can make use of this

[...]

PR #360, opened 2017-08-30
Refactor predicates into model

tiltec:
I thought it makes sense to have the status checks (=predicates) inside the model.

Probably also helpful to have them at one place if we are going to use django-rules #353


foodsaving/groups/models.py
```diff
@@ -44,3 +44,6 @@ def add_member(self, user, history_payload=None):
     def remove_member(self, user):
         pre_group_leave.send(sender=self.__class__, group=self, user=user)
         self.members.remove(user)
+
+    def is_member(self, user):
+        return user in self.members.all()
```
nicksellen: `self.members.filter(user=user).exists()`? Otherwise it's instantiating model objects for all the users (unless python/django allows something fancy)

nicksellen:
Looks good to me (with minor inline comment).

foodsaving/groups/models.py
nicksellen: .. oh, forgot to mark it as review comment. This one is now.

PR #362, opened 2017-08-31
Remove unused stuff

nicksellen:
Just a bit of a tidy up of some (hopefully) unused code. tests still work!


foodsaving/utils/mixins.py
```diff
@@ -15,11 +15,4 @@ def partial_update(self, request, *args, **kwargs):
         serializer = self.get_serializer(instance, data=request.data, partial=partial)
         serializer.is_valid(raise_exception=True)
         self.perform_update(serializer)
-
-        if getattr(instance, '_prefetched_objects_cache', None):
```
tiltec: This is copied 1:1 from `UpdateModelMixin`: https://github.com/encode/django-rest-framework/blob/27c382c98d0096cdf326313c8d8af17fce445a24/rest_framework/mixins.py#L72

Should probably be kept in case we might use `prefetch_related`.

Ideally, we wouldn't need to copy/paste it, but I needed to get rid of the `update` method in the mixin to disable `PUT` requests. Would be happy about other solutions.

nicksellen: It's not under any test coverage currently, 3 options:
1. remove it
2. add a test that makes use of it
3. find a way to call the rest_framework method without inheriting from it

I think 3 is best. 2 would be odd as the test would be testing something not used in the code. 1 would presumably make the code break in the future, and we would struggle to see what is happening.

Issue #230, opened 2016-12-25
Use django-configurations to manage settings.py


tiltec:
Config is getting more complex, so there's https://django-configurations.readthedocs.io/en/latest/ to help us.

nicksellen:
Now I think is a good time to revive this, I actually needed a dev dependency that is used in the tests.

I tried an experiment so load config from `local_settings_d` directory, by adding this to `settings.py`:

```py
import os.path
import glob

local_settings_files = glob.glob(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'local_settings_d', '*.py'))
local_settings_files.sort()
for file in local_settings_files:
    module_name = '{}.local_settings_d.{}'.format(os.path.basename(os.path.dirname(os.path.abspath(__file__))), os.path.basename(file)[:-3])
    module = __import__(module_name, globals=globals(), locals=locals(), fromlist=('*'), level=0)
    for setting in dir(module):
        if setting == setting.upper():
            locals()[setting] = getattr(module, setting)
```

However, I think an environment-based approach is probably better.

The other main problem right now is the docker-compose setup mounts a `local_settings.py` file into the right place, but you IDE does not see it (as it's inside the docker container). Then when you edit `local_settings.py` it writes the file and that shadows the mounted one, and then it no longer has the magic docker settings, so tests don't run any more etc.

I'd like the docker-specific settings to have just the things that are needed differently. Perhaps having the docker setup use a different `DJANGO_SETTINGS_MODULE` would be better approach, to make it more explicit, maybe `docker_compose_settings.py`.

tiltec:
I'd close this, it seems like we are doing fine juggling `local_settings.py` files...

PR #366, opened 2017-08-31
allow to run tests without FCM_SERVER_KEY

tiltec:
But it would be better to solve this with a separate test config #230

### DIFF
--- a/foodsaving/subscriptions/tests/test_fcm.py
+++ b/foodsaving/subscriptions/tests/test_fcm.py
@@ -19,30 +19,28 @@ def logger_warning_mock():
 
 
 @contextmanager
-def fcm_reload_without_config():
+def override_fcm_key(key=None):
     from django.conf import settings
-    if not hasattr(settings, 'FCM_SERVER_KEY'):
-        reload(foodsaving.subscriptions.fcm)
-        yield
-    else:
-        original = settings.FCM_SERVER_KEY
-        del settings.FCM_SERVER_KEY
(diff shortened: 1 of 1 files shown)


tiltec:
ooooh, it hurts test coverage!

foodsaving/subscriptions/tests/test_fcm.py
```diff
@@ -21,42 +21,60 @@ def logger_warning_mock():
...
         del settings.FCM_SERVER_KEY
         reload(foodsaving.subscriptions.fcm)
         yield
-    finally:
         settings.FCM_SERVER_KEY = original
         reload(foodsaving.subscriptions.fcm)
 
 
+@contextmanager
+def fake_fcm_key():
```
nicksellen: Maybe it can be combined with `fcm_reload_without_config` to be one function that supports bot `with fcm_key(unset=True)` and `with fcm_key(value='foo')`

tiltec:
Changed it @nicksellen - how does it look now?

foodsaving/subscriptions/tests/test_fcm.py
```diff
@@ -19,30 +19,28 @@ def logger_warning_mock():
 
 
 @contextmanager
-def fcm_reload_without_config():
+def override_fcm_key(key=None):
```
nicksellen: I suggested the `unset` option so you can actually unset it, passing the value as `None` is slightly different. But I think it's a tiny detail and nobody cares. So, I'm happy!

PR #367, opened 2017-08-31
Minor changes/fixes for feedback model

tiltec:
- fixed some typos and removed comments
- removed 'given_by' when creating feedback, as it's supplied by the serializer
- fixed bug in test_create_future_feedback_fails_as_collector and made the test work
- allow empty weight and comment values (for now, should enforce at least weight or comment)

(also added tblib to the dev requirement for getting tracebacks when running tests in parallel)


foodsaving/stores/tests/test_feedback_api.py
```diff
@@ -165,5 +182,6 @@ def test_create_future_feedback_fails_as_collector(self):
         Collector is NOT allowed to leave feedback for future pickup
         """
         self.client.force_login(user=self.collector)
-        response = self.client.post(self.url, self.future_feedback_post, format='json')
+        response = self.client.post(self.url, self.future_feedback_post)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST, response.data)
+        self.assertEqual(response.data, {'about': ['The pickup is not done yet']})
```
tiltec: Added the check, @mddemarie

foodsaving/stores/models.py
```diff
@@ -34,7 +34,7 @@ class Feedback(BaseModel):
     given_by = models.ForeignKey('users.User', on_delete=models.CASCADE, related_name='feedback')
     about = models.ForeignKey('PickupDate')
     weight = models.PositiveIntegerField(blank=True, null=True)
-    comment = models.CharField(max_length=settings.DESCRIPTION_MAX_LENGTH)
+    comment = models.CharField(max_length=settings.DESCRIPTION_MAX_LENGTH, blank=True)
```
tiltec: @id-gue I couldn't submit a feedback without comment through the API. It seems necessary to set `blank=True`, per https://docs.djangoproject.com/en/1.11/ref/models/fields/#blank

I undid the changes to `weight`, it seems necessary to have both `blank` and `null` there: to allow for empty input and to store empty input as null/none.

PR #375, opened 2017-09-03
Add group membership roles

nicksellen:
Fixes https://github.com/yunity/foodsaving-backend/issues/373

First pass at implementing group roles as described in https://github.com/yunity/foodsaving-frontend/issues/550.

There was some feeling to move away from the concept of "admin" roles, and the important concept was just the role to let people modify the group agreement.

However, this missed out the most important concept that we need a way to allow those roles to be assigned in the first place, so there are two roles to think about:
1. managing the member roles
2. editing the group agreement

This PR just implements the first one of those, so:
* added a postgres `ArrayField` to `GroupMembership` model to contain a list of strings
* add API endpoint `PATCH /api/groups/{groupid}/memberships/` which accepts user id, roles to add, roles to remove, returns the membership info (roles, created_at)
* added `memberships` field to the group detail info, it returns a _map_ of user -> membership info in addition to the existing `members` array of user ids (thinking: does not break existing API, allows easy way in to get membership info in frontend) - see below
* added `membership_manager` role (defined in roles.py)
* added code to bootstrap groups to have at least 1 of them if there is at least 1 member (per the original idea)

Group detail info (just notable fields):

```json
{
  "members": [
    1,
    6,
  ],
  "memberships": {
    "1": {
      "created_at": "2017-09-03T11:40:18.430772Z",
      "roles": [
        "membership_manager"
      ]
    },
    "6": {
      "created_at": "2017-09-03T11:40:20.604046Z",
      "roles": []
    }
  }
}
```


nicksellen:
I would add more tests if we like this approach, so mostly looking for @tiltec to see if he likes the general idea!

tiltec:
I had some thoughts about API style and it's implementation.

My preferred solution would be 
POST `/groups/{groupid}/memberships/{userid}/{membershipname}/`
DELETE `/groups/{groupid}/memberships/{userid}/{membershipname}/`

Advantages:
- one request per action
- easy to return success/failures status

Disadvantages:
- complexity in viewset

---

The API choice here is PATCH `/groups/{groupid}/membership/`.

Advantages: 

- only needs one detail_route in the viewset

Disadvantages: 

- different style than existing `/join/` and `/leave/` methods
- needs complex serializer (see comment there).
- hard to return specific errors if one role could be added, but not the other one

---

Another possibility: use PATCH `/groups/{groupid}/membership`, but instead have the whole set of roles `{user: 4, roles: ['role1', 'role2'] }` in the body (instead of `add_roles` and `remove_roles`)

Advantages: 

- similar data format for input and output
- simple serializer

Disadvantages: 

- different style than existing `/join/` and `/leave/` methods
- unwanted changes if users submit requests based on old data

tiltec:
image

Created a nested route with a `detail_route` and split up the serializers in two.

I tried to use DRF-Extensions nested routers, but it doesn't offer enough control when filtering querysets.

tiltec:
Anything to add?

nicksellen:
looks great! maybe split post and delete into two seperate methods, why combine? and I thought you were going to put add/remove into the model? and maybe try and see if the enums work out? I feel a bit mixed about the enums as the database no longer makes sense on its own, you need the code to know what the numbers mean. but not a strong feeling.

nicksellen:
also might be better as put rather than post. as the url already refers to the resource.

tiltec:
Moved the add/remove methods into the model.
Made it PUT rather than POST.

Bot endpoints have the same route. Having both endpoints in separate method breaks the routing, only the first one will ever get called.

I don't bother refactoring to enums, in fact I would be fine to complete remove enumfields from the project. I also see that advantage that the db makes sense by itself. can help debugging.

nicksellen:
Looks good for merge after a master rebase/merge :)

Issue #374, opened 2017-09-03
Rename "history" to "activity"


vinibiso:
Like issue [427](https://github.com/yunity/foodsaving-frontend/issues/427) from front-end, started by @tiltec.

Assigned to me.

tiltec:
Closing this because of no activity and because the frontend PR got out of date.

PR #377, opened 2017-09-04
rename history instances activity

vinibiso:
#374


tiltec:
Looks good! 

Unfortunately all the previous history items were gone after running the migrations on a existing database (with `./manage.py migrate`). It seems by renaming all migrations, you tell Django to create a completely new database table.

You can reproduce this by running these commands:

```py
# switch to master
git checkout master

# delete and recreate your local dev database
./manage.py reset_db
./manage.py migrate

# create some example data
./manage.py create_sample_data
# Note down one of the displayed email address, it allows you to login with password 123.
# For convenience, choose the second to last address . The last one is not part of any group.

# inspect the API response
./manage.py runserver
```

Now you can visit http://localhost:8000/api-auth/login/?next=/docs/ and login with one of those example accounts that you noted down before.
Clicking on "GET /api/history/" or http://localhost:8000/docs/#!/history/history_list and clicking the "Try it out" button should show you a list of history events.

```json
{
  "count": 45,
  "next": null,
  "previous": null,
  "results": [
    {
      "id": 224,
      "date": "2017-09-04T13:28:36.904713Z",
      "typus": "STORE_DELETE",
      "group": 5,
      "store": 21,
      "users": [
        30
      ],
      "payload": null
    },

```

Now let's try out the refactored version.

```py
git checkout rename_history_to_activity
./manage.py migrate
./manage.py runserver
```

Now http://localhost:8000/docs/#!/activity/activity_list shows an empty list: 

```json
{
  "count": 0,
  "next": null,
  "previous": null,
  "results": []
}
```

However, the content should be the same as before. 

How can we proceed while keeping existing data? Renaming migrations seems to be the wrong way to rename a Django app and model.

There is the [RenameModel](https://docs.djangoproject.com/en/1.11/ref/migration-operations/#renamemodel) migration command, maybe that could be helpful? It says you need to create an empty migration and add the command.

The next thing would be to find out how a Django app can be renamed. This seems tricky: https://stackoverflow.com/questions/8408046/how-to-change-the-name-of-a-django-app (though some of the answers are for older versions of Django, we use 1.11)

Maybe the simplest solution would be
- create a new app called activity, copy over all code from history and run `manage.py makemigrations`
- add a [data migration](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations) to move all database content from `history` to `activity`
- delete the history model and run `manage.py makemigrations` a second time

What do you think? It seems quite complex, but could be nice to solve in a clean way.

vinibiso:
So yeah. Here's what's happening.

Because the actual migration files have changes, django understand that and has to actually run them again. Which is horrible.
That makes it so that, it actually creates a whole new table name history, but because the app changed it becomes acitivity_history. And because it's new, it's actually empty.
So when I make the renaming migration it renames activity_history to activity_activity.
What I'm gonna try to do is actually go inside DB delete activity_history and activity_history_users and then rename history_history and history_history_users to activity_activity and activity_activity_users.

vinibiso:
I believe that solves it.

I created a empty migration and made some custom raw sql instruction to make everything work.

It takes in consideration both table that were the problem. 

I tested when coming from an "old" version and from a new one. Everything seems fine as far as the records in my tests.

tiltec:
I'm trying to understand the consequences and side effects of the SQL instructions. Will it also work if no `history_history` table is present, e.g. when initializing a new database?

Can you add a migration test, similar to [this one](https://github.com/yunity/foodsaving-backend/blob/a9f2c2a848f03960652a2eb7c2bd14ddb4263453/foodsaving/users/tests/test_data_migrations.py)? Then I'd feel more comfortable with merging.

tiltec:
Sorry to introduce so many conflicts here... I made changes to our signals today (basically removed custom signals).

PR #387, opened 2017-09-07
Add validations to the Feedback to check if there is valid content

id-gue:


id-gue:
Hi @tiltec 

could you have a look on our last commit? We are working on the PATCH endpoint at the moment and we thought it might be a problem if every collector can change every feedback. Do you think a new permission is a good way to make sure that everyone can change only their own messages? 

Thank you!

tiltec:
Adding a permission class depending on the request method (PATCH) sounds like a good solution!

PR #388, opened 2017-09-07
Keep track of which messages a user has seen

nicksellen:
This will be useful at some point for recording which messages in a conversation the user has seen.

* adds a `seen_up_to` field on the conversation/user join table
* does not track each message, assumes if you've seen message 10, you've seen 0-9 too
* _seen_ rather than _read_ to make a slightly less strong claim than seen...
* one API method to update `PATCH /api/conversations/{convid}/participant/` with data `{ "seen_up_to": <some messaage id> }`
* the `/participant/` part meaning the participant resource corresponding to the requesting user
* adds `seen` field for messages API

I also wanted to add it now, as it will inform some of the decision about the implementation of comment threads (as to operate it in a slack-like way, we'd need to track seen messages for conversation/user pairs and conversation/thread pairs separately - could mean threads should be their own conversation in the database... or some other options).

Todo:

- [x] get feedback on proposal
- [x] add tests


tiltec:
What are the use-cases of a `seen` field?

nicksellen:
Use cases:
* make conversations that have unread messages stand out from the others
* notify users about unread messages (e.g. daily digest, or individual message notifications)

tiltec:
I think that's a simple solution that should work for most cases.

Regarding threads: maybe it's easy enough to introduce "conversations-in-conversation"? Subscription handling could get a problem.

nicksellen:
Yes, I agree. So current changes are:
- expand conversations read api
  - `GET /conversations/`
  - `GET /conversations/{id}/`
- keep `seen_up_to` as a read only field on the conversation serializer
- add `POST /conversations/{id}/mark/ { seen_up_to: <id> }` endpoint

to do in the future:
- remove `GET /groups/{id}/conversation/` endpoint
- return conversation info about the thing the conversation is about

nicksellen:
Ok, also added `unread_message_count`. It assumes the message ids are ever increasing, which is the default for postgres seq's (but did not check specifically that we use those).

> The CYCLE option allows the sequence to wrap around when the maxvalue or minvalue has been reached by an ascending or descending sequence respectively. If the limit is reached, the next number generated will be the minvalue or maxvalue, respectively.
> 
> If NO CYCLE is specified, any calls to nextval after the sequence has reached its maximum value will return an error. If neither CYCLE or NO CYCLE are specified, NO CYCLE is the default.


https://www.postgresql.org/docs/9.1/static/sql-createsequence.html

PR #389, opened 2017-09-07
Add channel presence

nicksellen:
Currently when sending the push notifications it will send them regardless of whether the user is online and has the conversation open. In the case the user is on their laptop chatting away, their phone would be getting the push notifications even though they are active on the chat.

The solution is chat presence - this PR adds the backend support, frontend support would be via something like https://github.com/defunctzombie/away.

New behaviour is that if the user has an active (i.e. non-away) channel subscription, they will not receive push messages.


tiltec:
Looks good, an improvement for sure.

I expect some messages in the transition period between "being away" and "detected as away" will get undetected, as they get sent as websocket notification and not as push notification. (imagine sudden disconnect, or getting up and going out of the house while leaving your computer on). So ideally we would introduce an asynchronous task to send messages as push if the user is away but hasn't read the messages yet. Related to #388

nicksellen:
> sudden disconnect

If the disconnect happens properly (`websocket.disconnect` channel event is received), then the `ChannelSubscription` item will be removed, otherwise it will hang around for timeout period (currently 5 minutes, although we haven't configured that to happen yet).

Once/if #388 is implemented/merged then then we would know which messages were not read, even if the disconnect was unclean.

> getting up and going out of the house while leaving your computer on

There is no way to distinguish that from the user just viewing another tab on the computer. If using the js away lib, the timeout could be configured to whatever seems sensible.

PR #401, opened 2017-09-23
switch history to cursor-pagination

tiltec:
better efficiency and stables results compared than limit-offset pagination

Same style as proposed for messages #400 

Should be compatible with angularjs frontend.

### DIFF
--- a/foodsaving/history/api.py
+++ b/foodsaving/history/api.py
@@ -1,6 +1,6 @@
 from rest_framework import viewsets
 from rest_framework.filters import DjangoFilterBackend
-from rest_framework.pagination import LimitOffsetPagination
+from rest_framework.pagination import CursorPagination
 from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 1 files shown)


foodsaving/history/api.py
```diff
@@ -1,16 +1,17 @@
...
 
 from foodsaving.history.filters import HistoryFilter
 from foodsaving.history.models import History
 from foodsaving.history.serializers import HistorySerializer
 
 
-class HistoryPagination(LimitOffsetPagination):
-    default_limit = 50
-    max_limit = 1000
+class HistoryPagination(CursorPagination):
```
nicksellen: Will this work with the existing frontend? ... getting the first page would be enough I guess

tiltec: Yes, it should work without problems. The frontend uses the provided URL in the `next` field of the pagination header.

PR #406, opened 2017-09-28
Feedback limitations: only one feedback per user per pickup

id-gue:


foodsaving/stores/serializers.py
```diff
@@ -245,6 +245,13 @@ class Meta:
 
     def create(self, validated_data):
         validated_data['given_by'] = self.context['request'].user
+
+        existing_feedback = len(FeedbackModel.objects.filter(
+            about=validated_data['about'], given_by=validated_data['given_by']
+        ))
+
+        if existing_feedback != 0:
+            raise serializers.ValidationError({'non_field_errors': [_('You already gave feedback for this pickup')]})
```
tiltec: Three remarks about this:

If you put this validation in a `validate` method and raise the ValidationError, you don't need to specify the non_field_errors field explicitly.

Using the `.exists()` queryset method should simplify things more: `if FeedbackModel.objects.filter(about=validated_data['about'], given_by=validated_data['given_by']).exists()`

Finally, using the DRF [UniqueTogetherValidator](http://www.django-rest-framework.org/api-guide/validators/#uniquetogethervalidator) seems to be the best choice here. Did you have a look at it yet?
(With this you don't need to care about the first two remarks anymore..)

mddemarie: Thank you very much for your feedback.

We have to do the setup again on our personal laptops first. We will get back to solving this this week. We can replace our `non_field_errors` with `.exists()`.

To the last topic about UniqueTogether Validation: we already tried that but then we replaced it with unique fields in our Feedback model. Can we rather use unique fields in the `FeedbackModel` instead?

tiltec: How are you doing? 

Sorry I didn't answer the last question, it's not clear to me what you mean with "...we replaced it with unique fields in our Feedback model. Can we rather use unique fields in the FeedbackModel instead?" That sounds exactly the same?

PR #407, opened 2017-09-28
Feedback limitations: limit the days to give feedback

id-gue:
We added a validator to make sure that collectors can give feedback only in the first 30 days after the pickup. All tests in the stores app work – but a test in invitations returns this error:

  File "/karrot-backend/foodsaving/invitations/tests/test_api.py", line 150, in test_list_invitations
    self.assertEqual(len(response.data), 1)
AssertionError: 0 != 1

Any ideas why? There shouldn't be a connection...


tiltec:
Seems fine, no test failures here or in the CI!

tiltec:
I ran the test and everything seems fine, so will merge.

Issue #409, opened 2017-10-03
Make (serializer) validation safe for concurrent requests


tiltec:
(Almost) all our serializer validations behave incorrect on concurrent requests, which leads to server errors or corrupted data.

For example joining a pick-up as collector: if there is one slot free and two users join at the same time, they might both pass the serializer validation. The database doesn't check for `max_collectors`, hence both will be added as collectors.

The reason for this behaviour is that both requests run in separate threads (EDIT: or processes even). What I usually imagine when writing backend code is the following:

```
_Request 1:_ start1 - read1 - validate1 - write1 - respond1
_Request 2:_ ------------------------------------------- start1 - read2 - validate2 - (fail validation!)
```

Request 2 fails because `read2` happens after `write1`.

Here's what supposedly happens when two requests arrive in short succession:

```
_Request 1:_ start1 - read1 ----- validate1 ---- write1 - respond1
_Request 2:_ --------- start2 - read2 ---- validate2 ---- write2 - respond2
```

Notice how `read2` and `validate2` happen before `write1`? Of course both requests will pass validation!

I experienced similar problems if the same user fires two request to join a pick-up in short succession. Here however, the database checks the `unique` constraint when writing and hence Django responds with HTTP500.

One fits-all solution is to wrap `create` and `update` ViewSet methods into Djangos [`@transaction.atomic`](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#django.db.transaction.atomic) decorator. This has a slight performance drawback, but should prevent all incorrect behaviour.

There are two caveats mentions in the [docs](https://docs.djangoproject.com/en/1.11/topics/db/transactions/#django.db.transaction.atomic): "Don't catch database exceptions inside atomic()" and "Revert model state manually when rolling back a transaction". We have to see how they apply.

tiltec:
Another solution would be to express all constraints at database-level and handle `IntegrityErrors` properly. But since this has some problems, it could only be used in addition to serializer validation and viewset permission handling.

More details:

### Django doesn't support custom constraints nicely (yet)

There is an active but long-running PR to add DB constraints via model `Meta`: https://github.com/django/django/pull/7615
It might arrive in Django 2.0, to be expected in December 2017.
Here's an example how to use it: https://github.com/Ian-Foote/django/blob/f416afe15933432488ee2c61ccea2dcef6ffa543/docs/ref/models/options.txt#L434

[...]

PR #411, opened 2017-10-15
WIP: add concurrency tests and database locks

tiltec:
Related to #409 
currently has unhandled deadlocks, needs more work or another approach


tiltec:
@nicksellen 
What's your opinion on this PR?

nicksellen:
Maybe we should close this one now? It's around 2.5 years old... I think the underlying issue is actually still very interesting and probably relevant for a lot of web apps, but also quite difficult to address ourselves with the current level of framework support (or lack of it).

tiltec:
I'll have another look at it soon, I actually have it in mind from time to time when one of the weirder backend errors come up in Sentry.

I might apply some locking similar as in this PR, but not add the test. It also might need some frontend changes, otherwise it might be confused and show a failed request, whereas the user just clicked the button twice very fast. Actually, most of the backend errors should go away when buttons are properly debounced…

In any case, I'll close this one soon!

nicksellen:
Yup, seems fine to be open if it's still active in your mind :)

PR #414, opened 2017-10-27
added status to store

alangecker:


foodsaving/stores/serializers.py
```diff
@@ -236,6 +236,11 @@ def validate_weeks_in_advance(self, w):
             raise serializers.ValidationError(_('Set at least one week in advance'))
         return w
 
+    def validate_status(self, status):
+        if status in ['created', 'negotiating', 'active', 'declined', 'archived']:
```
tiltec: We better use a ChoiceField in the serializer, like here: 

https://github.com/yunity/karrot-backend/blob/c09047955d1bbec7e476beea4197ee5c3da34353/foodsaving/groups/serializers.py#L168-L172

And define the statuses as constant, e.g. in a class.

Issue #385, opened 2017-09-07
Add optional agreement for groups


nicksellen:
This is the backend implementation of https://github.com/yunity/foodsaving-frontend/issues/324#issuecomment-312685256, this is the proposal from a discussion between me, @tiltec, and @djahnie:

> 0. The legal agreement should be optional and per-group, decided by group admins
> 1. We need some people (group admins) in the group who are able to edit the legal agreement -> #356 
> 2. If the legal agreements gets changed, all members have to accept again (simpler implementation)
> 3.1 If users are already in the group and have not accepted the latest agreement, they should be showed a dialog and be prompted to agree on it.
> 3.2 If users join a group, joining and accepting should be done in one step
> 
> Probably we could record the date when the users agreed, so we could later compare it to the change time of the agreement.

I don't think any of the subsequent discussion changes that idea.

For implementation, my initial idea was to:
* add a new group role (using our new roles system) called `GROUP_AGREEMENT_MANAGER` (the existing one being `GROUP_MEMBERSHIP_MANAGER`)
* add a `groups.Agreement` model which has `title`, and `content` text fields
* add a one2one relationship from group -> agreements (denoting the active agreement)
* add a foreign from agreements -> group (denoting which group an agreement is related to)
* add an agreed_to foreign key to agreements on group membership (recording the agreement a given user has agreed to)

Rationales:
* minor changes to the agreement (e.g. typo, italic change, new line added) does not need to get everyone to re-agree (what is minor is up to the group/law to decide)
* we record explicitly which agreement a user agreed to instead of implicitly by date
* we keep the history of agreements as is useful (could prune them where nothing refers to them anymore, but seems potentially useful to keep any agreement that is refered to in an agreed_to field, presumably the old agreement is still fine for letting them browse the site, but they would need to accept the new agreement to do certain other things)

tiltec:
Seems to be a good model. I thought about using something like https://github.com/etianen/django-reversion from model versions & rollback, but it doesn't fit this purpose.

How would the transition between "group without agreement" and "group with agreement" look like? And the reverse?

**adding agreement**

- the user with `GROUP_MEMBERSHIP_MANAGER` appoints someone as `GROUP_AGREEMENT_MANAGER`
- `GROUP_AGREEMENT_MANAGER` pushes an agreement and set it as active (one2one)
- all users have to agree before they can do any action inside the group

**removing an agreement***

- `GROUP_AGREEMENT_MANAGER` removes the active agreement

[...]

PR #417, opened 2017-10-28
Add group agreements

nicksellen:
Implementation of https://github.com/yunity/karrot-backend/issues/385

A few things missing so far:
- try and reduce the duplication in my serializers
- ensure the membership manager can assign agreement managers
- probably initialize an agreement manager so there is an initial one
- limiting certain actions depending on agreement status (e.g. users should not be able to do pickups unless they have agreed)

Two things I think I don't need to implement (at least right now) are:
- deleting agreements
- unagreeing (gets more complicated as it might depend on what is inside the agreement content, so might need to be set whether it can be unagreed or not, per agreement, probably does not matter for now)


foodsaving/groups/api.py
```diff
@@ -144,3 +145,37 @@ def modify_user_roles(self, request, pk, user_id, role_name):
...
 
         return Response(GroupMembershipInfoSerializer(instance).data)
+
+
+class IsGroupAgreementManager(BasePermission):
+    message = _('You cannot manage agreements.')
+
+    def has_permission(self, request, view):
+        group_id = request.GET.get('group')
+        print('checking agreement manager permissions', request.user, group_id)
```
tiltec: remove print before merging

foodsaving/groups/serializers.py
```diff
@@ -108,6 +119,64 @@ def create(self, validated_data):
         return group
 
 
+class AgreementSerializer(serializers.ModelSerializer):
```
tiltec: Shouldn't this also return the users that have agreed? Can be shown in frontend to the agreement manager.

nicksellen: Could do, and might be useful at some point, I just implemented the "minimum information" principle for now...

Another approach would be to include it in the membership objects in the group serializer (for membership managers), as that is where other membership management information is (e.g. roles).

tiltec: Yes, that seems to be easy enough to get added later when the frontend is in development or when a user requests it. It's fine to progress without this information for now.

PR #427, opened 2017-11-04
Clean up users API

tiltec:
The `/api/users/` should be uniformly used to retrieve non-private user information. This PR adds a new `/api/auth/user/` endpoint to deal with getting, updating, creating and deleting a (logged in) user account.
A follow-up PR will remove this functionality from the `/api/users/` endpoint, after the front-end has been migrated.

To do:

- [ ] maybe move verify-mail, resend-verification also into auth-user?
- [ ] reset-password could get its own endpoint probably


tiltec:
(Codecov reports wrongly, has full coverage)

nicksellen:
Is this waiting for your TODOs to be done?

tiltec:
yes. the todos are more like "evaluate if it makes sense", so could also be merged as it is

PR #435, opened 2017-11-13
Extend README.md

mr-kojo:
Add some words regarding Django migrations

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ source env/bin/activate
 ./manage.py migrate
 ```
 
-The very first migration initializes your database. Thus, you will need to execute the above commands before you launch your backend for the first time.
+In particular, before you launch your backend for the very first time you will need to execute `./manage.py migrate` to initialize your database.
 
 # Contributing to karrot-backend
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -156,14 +156,16 @@ You can run the server with `python manage.py runserver`.
...
-Sometimes you will need to create Django migrations.
+Sometimes you will need to create Django migrations. 
 
 ```
 source env/bin/activate
 ./manage.py makemigrations
 ./manage.py migrate
 ```
 
+The very first migration initializes your database. Thus, you will need to execute the above commands before you launch your backend for the first time.
```
tiltec: It's enough to run `./manage.py migrate`, if `./manage.py makemigrations` does something after a fresh `git clone`, somebody forgot to do it before committing (which usually breaks thing in deployment).

`./manage.py makemigrations` goes through your code and detects if you changed a model. If yes, it will create a new migration.

mr-kojo: Yes, I got that :)
But you're right, I haven't pointed it out clearly enough. Now it should be better.

PR #437, opened 2017-11-14
Add user photo field using versatileimagefield

nicksellen:
The upload worked but I couldn't actually see how to serve the files, the provided URLs did not work directly.


tiltec:
Does it dump them into `uploads/user__photos/`?

nicksellen:
Yes, maybe need to create the dir first (I tried it with the `uploads/` prefix and it created `__sized__` and `user__photos` dirs at the top level).

The urls it returned were like `/api/auth/user/__sized__/blahblhahblahbla/oijsef/bouasef.jpg` which our `urls.py` doesn't seem to have any clue about. The docs didn't seem to help me out either.

nicksellen:
You can use curl to upload them, something like:

```
curl -XPATCH localhost:8000/api/auth/user/ -F photo=@someimage.jpg
``

tiltec:
VersatileImageField extends Django's `ImageField`, I found the documentation here: https://docs.djangoproject.com/en/2.0/topics/files/

I'll work on this PR.

tiltec:
@nicksellen 
You wanna have a look at it before I merge?

tiltec:
The default storage backend is [FileSystemStorage](https://docs.djangoproject.com/en/2.0/ref/files/storage/#the-filesystemstorage-class), which dumps the files in the `MEDIA_ROOT` folder.

tiltec:
Could use some tests too...

tiltec:
There are security considerations when serving user-uploaded content from your own domain: https://docs.djangoproject.com/en/2.0/topics/security/#user-uploaded-content-security

We could serve the images from a separate domain (e.g. usercontent.foodsaving.world) or use some CDN.

PR #440, opened 2017-11-17
Support websocket auth using token

### LINKED ISSUE #439: Get websocket user from token
When using token auth, channels does not make the user available.

A few ideas for implementing something:
- https://github.com/django/channels/issues/510#issuecomment-288677354
- https://gist.github.com/leonardoo/9574251b3c7eefccd84fc38905110ce4
- https://gist.github.com/leonardoo/9574251b3c7eefccd84fc38905110ce4

Frontend task https://github.com/yunity/karrot-frontend/issues/766

nicksellen:
Seems a better approach is (ab)using websocket subprotocols to pass the token, some other people doing that:

- https://stackoverflow.com/a/35108078
- https://trello.com/c/lRdCiGzd/1026-2-authenticate-websocket-api-requests-using-subprotocols-not-query-parameters
- https://stackoverflow.com/questions/36323743/oauth2-authentication-for-websockets-pass-bearer-token-via-subprotocols

This is all because you cannot set custom headers for websocket connections, only use this protocol argument (`new WebSocket(url, protocols)`), or query params on the url.

It works like this:
- frontend passes two security protocol entries:
  1. `karrot.token`
  2. `karrot.token.value.<base64token>` *
- backend accepts `karrot.token` as a valid protocol (via `CHANNELS_WS_PROTOCOLS` option)
- when a websocket connection comes in pull the token out from the second header
- we authenticate the user, but also set the token on the channels session
- during a subsequent websocket request we get the token and authenticate again
- if no protocols are set, then the session based auth still works as it did before

\* _base64 padding (`=` at the end) was not accepted as a valid ws protocol value by chrome, so I stripped them out, and added them back in on the server, a bit hacky, but means we can support any token, although our current tokens would be fine as they are I think, so could remove this base64 stuff_

Logout will not clear the channels session, I think it will stay around as long as the socket is open. Our client will disconnect the websocket on logout.

Fixes #439

### DIFF
--- a/foodsaving/subscriptions/consumers.py
+++ b/foodsaving/subscriptions/consumers.py
@@ -1,7 +1,37 @@
+from base64 import b64decode
+from urllib.parse import unquote
+
 from channels.generic.websockets import JsonWebsocketConsumer
 from django.utils import timezone
+from rest_framework.authentication import TokenAuthentication
 
 from foodsaving.subscriptions.models import ChannelSubscription
+
+token_auth = TokenAuthentication()
+
+
+def check_for_auth_token_header(message):
+    prefix = 'karrot.token.value.'
+    for header, value in message['headers']:
+        if header == b'sec-websocket-protocol':
+            protocols = [x.strip() for x in unquote(value.decode('ascii')).split(",")]
+            for protocol in protocols:
+                if protocol.startswith(prefix):
+                    value = protocol[len(prefix):]
+                    if len(value) % 4:
+                        # not a multiple of 4, add padding:
+                        value += '=' * (4 - len(value) % 4)
+                    token = b64decode(value).decode('ascii')
+                    message.channel_session['auth_token'] = token
+
+
+def check_for_token_user(message):
+    if not message.user or message.user.is_anonymous():
+        auth_token = message.channel_session.get('auth_token', None)
+        if auth_token:
+            user, _ = token_auth.authenticate_credentials(auth_token)
+            if user:
+                message.user = user
 
 
 class Consumer(JsonWebsocketConsumer):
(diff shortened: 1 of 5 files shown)


nicksellen:
Oh, my fancy new f-strings aren't supported in the test environment :(

They were introduced in python 3.6, oh, we only have `Python 3.5.3` on the server, I guess I need to remove them again.

>    message_title=f'{message.author.display_name}: {message.content}',                          ^
> SyntaxError: invalid syntaxSyntaxError: invalid syntax

nicksellen:
Oh, codecov is so mean!

foodsaving/subscriptions/consumers.py
```diff
@@ -1,31 +1,64 @@
...
+
+def check_for_auth_token_header(message):
+    prefix = 'karrot.token.value.'
+    for header, value in message['headers']:
+        if header == b'sec-websocket-protocol':
+            protocols = [x.strip() for x in unquote(value.decode('ascii')).split(",")]
+            for protocol in protocols:
+                if protocol.startswith(prefix):
+                    value = protocol[len(prefix):]
+                    if len(value) % 4:
```
tiltec: Does this happen in normal operation or is it to protect against attacks? In any way, would be good to cover the other case (multiple of 4) in a test case as well.

nicksellen: Normal case, e.g. `foo` base64's to `Zm9v` (len=4), and `foo1` to `Zm9vMQ==` (len=8). The other case is standard base64, so less interesting, but I'll add a test anyhow.

nicksellen: Test added!

PR #442, opened 2017-11-17
Better dev dependency handling

### LINKED ISSUE #355: Better support for dev dependencies
I added a dev library for use in testing, then found the workflow for managing dev dependencies a bit lacking.

Currently for dependencies I run `pip-compile && pip-sync`, but this would uninstall things from `dependencies-dev.txt`.

https://github.com/jazzband/pip-tools/issues/398 has some ideas. I'm thinking to include a `dependencies-dev.in`.

Then updating everything is done with:
```
pip-compile && pip-compile requirements-dev.in && pip-sync requirements*.txt
```

Which could go into an `update-deps.py` command (so it's cross platform).

nicksellen:
Fixes #355 

Adds two commands to the project root:
- `./compile.py` - runs pip-compile for requirements.in and requirements-dev.in
- `./sync.py` - runs pip-sync for requirements.txt and requirements-dev.txt

I tried out https://docs.pipenv.org/ and it's nice, but soooooooooooo slooooooowwwwwwwwwwwww compared to using pip-tools.

Also recompiled the deps which meant `text-unidecode` is now installed and the errors of it missing have gone away, so I guess we had incorrectly compiled deps before.


tiltec:
Ah, could be that pyup.io is less than stellar at keeping requirements.io intact!
Maybe we should also reference sync.py in the Readme…

Issue #430, opened 2017-11-07
Rework change-email functionality


tiltec:
We should have a look how other websites implement change-email and copy an approach that seems good.

Things to consider:

- handle initial verification different from verifying a changed e-mail address (user should stay verified)
- allow aborting change requests (currently not possible)
- prevent changing to an e-mail address that another user already uses (take case-insensitivity into accout, should currently work)

tiltec:
Moved to https://github.com/yunity/karrot-frontend/issues/927

PR #446, opened 2017-11-20
Simplify users and groups API

### LINKED ISSUE #370: Remove /api/groups/ overload
The swagger view for /api/groups/ is confusing, it only shows the fields that a non-member would see. But during creation, you can set more fields (like `password`, `description`).

https://dev.foodsaving.world/docs/#!/groups/groups_create

image

Maybe a good solution would be to _not_ overload the endpoint. `/api/groups/` should only handle groups that the user is a member of. An second `/api/groups-info/` endpoint could be created for retrieving and listing public group info.

tiltec:
Closes #370 

- Removes unauthenticated (public) information from `/api/groups/` -> now only at `/api/groups-info/`
- Removes authUser related stuff from `/api/users/` -> now at `/api/auth/user/`
    - also valid for `verify_mail` and `reset_password`
- `/api/users/resend_verification/` -> `/api/auth/resend_verification/`
- Removes `/api/auth/status/` endpoint -> use `/api/auth/user/` now
- Removes sensitive fields such as `key_expires_at` and `unverified_email` from user profiles (accessible via `/api/users/`)
- Allow changing e-mail address back to previous one (related to #430)
- Fixes spelling of `read_only_fields` in userauth view


tiltec:
There seems no need to nest the additional auth-user related actions so deep, e.g. `/api/auth/user/reset_password/` could also go to `/api/auth/reset_password/`.

tiltec:
Should be ready, looking forward to your review!

config/urls.py
```diff
@@ -27,31 +27,32 @@
...
+from foodsaving.userauth.api import AuthUserView, AuthView, LogoutView, VerifyMailView, ResendVerificationView, \
+    ResetPasswordView
 from foodsaving.users.api import UserViewSet
 
 router = routers.DefaultRouter()
 
-router.register(r'groups', GroupViewSet)
-router.register(r'groups-info', GroupInfoViewSet)
-router.register(r'agreements', AgreementViewSet)
+router.register('groups', GroupViewSet)
```
nicksellen: Oh, wonder what the `r` was for.

config/urls.py
```diff
@@ -65,9 +66,11 @@
 
 urlpatterns = [
     url(r'^api/auth/token/$', obtain_auth_token),
-    url(r'^api/auth/status/$', AuthStatusView.as_view()),
     url(r'^api/auth/logout/$', LogoutView.as_view()),
     url(r'^api/auth/user/$', AuthUserView.as_view()),
+    url(r'^api/auth/user/verify_mail/$', VerifyMailView.as_view()),
+    url(r'^api/auth/user/resend_verification/$', ResendVerificationView.as_view()),
```
nicksellen: I would agree with your idea to make it less nested (which also makes it more consistent with `reset_password`.

tiltec:
It should be deployed at the same time as the frontend patch, because of the modifications to `verify_mail`, `reset_password` and `resend_verification`.

config/urls.py
tiltec: `r` makes it a [raw literal](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals), which commonly used for specifying regular expressions. Not needed here, so I removed it.

(`r` actually not _needed_ for regexp, but it's a [convenience method](https://stackoverflow.com/questions/3517802/python-raw-literal-string) to prevent the need for double escaping stuff)

config/urls.py
tiltec: I'll move them one level up.

Issue #368, opened 2017-09-01
Refactor: Split up store app


tiltec:
Currently the store app is the most complex Django app in our project. It contains `store`, `pickupDate`, `pickupDateSeries` and `Feedback` models, serializers, views and related logic.

It seems easily possible to split it up into separate Django apps to have better overview and less unrelated code when modifying files.
We could make it so that Django apps largely mirror API endpoint names.

nicksellen:
Will be especially helpful if we make pickup dates abstracted into tasks.

tiltec:
Correct @vinibiso, plus pickupDateSeries could also be a separate app. They are somehow coupled to the pickupDate model, but not very tightly.

Currently, there are two devs working on the Feedback model #253, so it would be good to wait until their changes are merged.

nicksellen:
I like the idea to separate into apps, but `pickupDateSeries` in it's own app is taking the concept too far IMO. The separation between app and model would become very fine. I'd probably just split `pickup*` and `feedback` models into a pickups app.

tiltec:
They all have dependencies to each other. Let me highlight some.

- `PickupDate` is dependent on  `Store`. The `series` ForeignKey is optional (only when it's part of a series).
- `PickupDateSeries` needs values from the `Store` model (`weeks_in_advance`) to generate `PickupDate` entries.
- `Feedback` depends on `PickupDate`.

So whatever we do, it means breaking things apart that are somehow coupled. Maybe it's best for now to do what Nick said, creating a new `pickups` app for PickupDate, PickupDateSeries and Feedback.

Maybe it makes sense to move the `weeks_in_advance` setting away from the store. It controls how many weeks in advance a series will create pickupDates (analog to foodsharing.de behavior). Maybe there's a better solution, e.g. controlling this setting per series and not per store.

PR #450, opened 2017-11-23
refactor stores app (#368)

mr-kojo:
- [X] Move models `PickupDate`, `PickupDateSeries`, `Feedback `to new `pickups` app
- [X] Create Django migrations such that no data get lost. (This was a bit tricky. I figured out how to do it with the help of [ozan's answer on StackOverflow](https://stackoverflow.com/questions/25648393/how-to-move-a-model-between-two-django-apps-django-1-7/26472482#26472482). See commit _Data migration_.)
- [x] Write migration test.


tiltec:
Looks great! Good work figuring out the `SeparateDatabaseAndState` migration, [it sounds powerful and dangerous](https://docs.djangoproject.com/en/1.11/ref/migration-operations/#separatedatabaseandstate).

Do you have an idea where the merge conflicts might come from? Just guessing, maybe #441?

I recommend to keep `weeks_in_advance` in the `Store` model for now, because moving it probably means changing the API and that would require frontend changes.

Instead, I would find it great if you could add a test for the migrations, because it will run on production data. @nicksellen implemented a migration test once [here](https://github.com/yunity/karrot-backend/blob/master/foodsaving/users/tests/test_data_migrations.py), maybe that's helpful. (Although I vaguely remember that it requires a reverse migration?)

mr-kojo:
Yes, the conflicts are probably due to #441. I guess the easiest way to resolve them is to take over my changes (which include not only code changes but also file movements) and then to apply the changes of #441 manually.

Yes, you are right, I should add a migration test.

Concerning `weeks_in_advance`, sooner or later we should really move it to `PickupDateSeries`, it does not make sense as an attribute of `Store` any longer, I think.

tiltec:
> Concerning weeks_in_advance, sooner or later we should really move it to PickupDateSeries, it does not make sense as an attribute of Store any longer, I think.

So far there wasn't any request from users to make this setting individually by series instead of once per store, do you think it makes sense to expose this functionality in the UI nevertheless?

mr-kojo:
> So far there wasn't any request from users to make this setting individually by series instead of once per store, do you think it makes sense to expose this functionality in the UI nevertheless?

I think it would actually be sufficient if the `weeks_in_advance` setting is adjustable per store. However, to me it seems that it is an attribute of a series and not one of a store: Every series needs to know this setting, but a store may not necessarily go along with a series (if it has only one-time pickups, or none at all). 

It's not high priority, of course, so I would suggest that we open a new issue _Move field weeks_in_advance to the PickupDateSeries model_.

tiltec:
> In any case, it's not high priority, of course. So how about opening a new issue Move field weeks_in_advance to the PickupDateSeries model?

Sounds good!

mr-kojo:
Created #451.

tiltec:
The test gives me enough confidence to release this into dev.karrot.world ;)

After you resolve the merge conflicts, we can merge this into master.

PR #456, opened 2017-12-07
Get ready for Django 2

tiltec:
Django 2 needs the `on_delete` attribute on foreign key constraints and `user.is_authenticated` is a property now instead of a function.


compile.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
 
 import subprocess
-subprocess.run(['pip-compile', 'requirements.in'])
-subprocess.run(['pip-compile', 'requirements-dev.in'])
+subprocess.run(['pip-compile', '-U', 'requirements.in'])
```
tiltec: I made this so that it actually updates the dependencies (if not pinned in `requirements.in`). I can't think of any case where I _just_ want to compile the deps without upgrading.

nicksellen:
Looks good. I read through most of it. But did not try it. I guess safe enough to try out from Dev.

Issue #275, opened 2017-03-06
Introduce token-based actions


tiltec:
The following actions should require a token (valid for single use, limited time, received via e-mail) to be performed:

- Password reset (currently sends out a permanent password)
- Account delete
- Mail verification (already done, but called activation key)
- Mail change (already done, but called activation key)

We could add a new `Token` model with the fields `token`, `valid_before`, `is_used`. Then we can reference it in the user model with one `OneToOneField` (see Django docs for more details) per purpose.

This has some advantages over adding all fields to the user model:

- separation of concerns
- more DRY (don't repeat yourself)
- concise naming (otherwise we would have names like: `passwordreset_token_valid_before`)

I think this issue could be split in these steps:

1. Discuss and create the `Token` model
2. Refactor mail verification and mail change to use the token model
3. Implement account delete with token
4. Implement password reset with token

(Side note: while 1 & 2 should work without frontend changes and 3 is not implemented yet in the frontend, 4 would need some changes there as well. That's why I put it last)

Josiah-DuBose:
Hello, I'd like to work on this issue if I could.

tiltec:
Hi, sounds great! Be sure to join our Slack on slackin.yunity.org for further discussion.

Josiah-DuBose:
is that slackin@yunity.org by chance?

djahnie:
Just visit [slackin.yunity.org](https://slackin.yunity.org/) and send yourself an invitation to join our team chat, @Josiah-DuBose. On the yunity Slack you'll find a channel named #foodsaving-tool, where all the discussions regarding development of the tool take place. :)

mr-kojo:
@tiltec suggested that we any term different from `Token` so that there is no confusion with [authentication tokens](http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication). I think that's a pretty reasonable idea so I made up my mind and came up with two candidates:

**_VerificationCode_**
PRO: Gives a pretty accurate idea of what it is meant for
CON: Some might think it solely represents an "e-mail verification code" although it can also stand for e.g. a "password reset verification code" (i.e., a code that verifies a requested password reset)

**_Key_**
PRO: Short
CON: Very general; _Key_ usually doesn't have the connotation that it can be used only once.

mr-kojo:
Assuming we call the new model _VerificationCode_, I would suggest the following fields:

- `verification_code`: The token
- `timestamp_created`: The time of creation
[...]

PR #463, opened 2017-12-22
Refactor e-mail verification to use VerificationCode model

mr-kojo:
Based on issue #275.

I'll continue right after the Christmas holidays 🎄


mr-kojo:
Meanwhile, I discovered [`django.contrib.auth.tokens.PasswordResetTokenGenerator`](https://github.com/django/django/blob/master/django/contrib/auth/tokens.py). This class creates and checks **non-persistent** password reset verification codes. Here, a verification code is simply a hash of the user object together with the current timestamp. Thus, when the user opens the password reset URL and enters a new password, the verification code gets invalidated automatically (because the user object changes and thus the hash changes, too). 
This mechanism could be used for email verification, email change, and account deletion just as well.

Now I wonder if I should implement it this way (or if I should stick to the solution with persistent verification codes).

PROS:
- Time-saving: Re-using code, no need to implement and maintain verification code functionality
- Non-persistent: No need to cram the database with (single-usage) verification codes
- Secure: Well-tested Django implementation using cryptographic hashes with salt

CONS:
- Less flexibility: This approach requires that the user object must not be modified by the time the user opens the verification URL. Hence, the user must not modify his/her profile in the meantime and she/he must not login again (because this would modify the `last_login` field). However, all of this could be enforced quite easily: We could simply log out the user and disable login until he/she opens the verification URL. 
- Less robust: If a user requests a verification mail and right afterwards an admin modifies the corresponding user for some reason, this will break the verification code. In such a case, the user will have to request a new verification mail.
- Harder to debug: Since verification codes are not stored persistently, there is no way to see e.g. if a user has requested one already and if so, if/when he/she has used it.

What are your opinions?

tiltec:
Nice find! I like solutions where we can avoid storing tokens in the database.

I think we can use the `PasswordResetTokenGenerator` in two different ways: for password reset directly and for other user-related tokens when subclassed.

In the subclass, I would modify theses bits:
- rename `settings.PASSWORD_RESET_TIMEOUT_DAYS` to something else to make it clear that it's not only used for password reset: https://github.com/django/django/blob/master/django/contrib/auth/tokens.py#L49
- remove dependency on user.last_login in `_make_hash_value()` 

Maybe something like this:
```
class TokenGenerator(PasswordResetTokenGenerator):
    def _make_hash_value(self, user, timestamp):
        return (str(user.pk) + user.password + str(timestamp))
```
[source](https://medium.com/@frfahim/django-registration-with-confirmation-email-bb5da011e4ef)

I'm unsure if we should also use a similar token generator for invitation codes, as I can't imagine a reliable source to calculate the hash from.

tiltec:
I had a look at [django-registration](https://django-registration.readthedocs.io/en/2.3/hmac.html), it calculates the hash only from the username field, adds a configurable salt to it and signs with Djangos `SECRET_KEY`:
https://github.com/ubernostrum/django-registration/blob/master/registration/backends/hmac/views.py#L64-L66

mr-kojo:
> I had a look at django-registration, it calculates the hash only from the username field, adds a configurable salt to it and signs with Djangos SECRET_KEY

Great, even more elegant! That's how I'll do it 👍

AUTHORS.md
```diff
@@ -3,6 +3,7 @@
 - [Douglas Webb](https://github.com/bbew/), dmhwebb@gmail.com
 - [Florian Gössele](https://github.com/goessek/), goessek@posteo.de
 - [Kristijan Miklobusec](https://github.com/Sliverriver), kristijan80@web.de
+- [Lukas Gebhard](https://github.com/mr-kojo), freerunningapps@gmail.com
```
tiltec: Oh the authors file! I think you're one of the few people listed here that contributed code to karrot ;)

mr-kojo: In particular, @nicksellen is missing

tiltec: Yes, I wonder why we need this file at all. Git history is much more useful.

mr-kojo:
I found two problems of my current implementation:

1)  A user can use the same verification code twice e.g. if he/she changes his/her email address twice within the time limit that is configured for email verification.
2) If a user changes his/her password and email at once and then opens the email verification URL, this will break the verification code for password change (because a verification code contains a hash of the email address).

I have been thinking about this for quite some time now and the best solution I found is this:

To solve (1):
- Before a verification code is generated, we ensure all previous ones have to be validated (used) already.
- Each verification code has an index number. The most recent index is stored in the database. When validating a verification code, we check if the index is lower than the one in the database. If so, the verification code is invalid because it has already been used before.

To solve (2):
We hash the user's primary key instead of the email address.

I am currently implementing this...
Let me know if you have concerns, though.

nicksellen:
I wonder if it's simpler to just put them in the database, explicit over implicit. I think it's easy to run into caveats and funny edge cases when doing clever things.

tiltec:
The fix for (1) looks like it would defeat the advantage of not having to store process-related data in the database. Maybe then it's better to save the code directly in the DB or what would speak against it?

mr-kojo:
You're right. Too bad, would've liked the non-persistent approach. 
Anyway, it was an interesting detour :)
Lesson learnt: Non-persistent verification codes are a clever way to authenticate actions that
- can only be done once and can't be undone (e.g., activate a non-editable username), or
- have to be done immediately (i.e., don't let the user do anything else before she/he has performed the action).

tiltec:
It's an interesting concept and it still seems suitable for activating the email/account. But now that we go for an explicit code implementation, maybe we'll also do email activation with it.

mr-kojo:
Ready to be merged (at least I hope so)

tiltec:
Thanks! I got the notification, will review tomorrow.
Is it compatible to the current frontend version?

foodsaving/users/migrations/0019_user_photo.py
```diff
@@ -7,13 +7,14 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('users', '0017_auto_20171207_1210'),
+        ('users', '0018_remove_verification_codes'),
```
tiltec: I think you must put new migrations after old migrations that already ran. I fear some problems on staging and master databases when inserting a new migrations between two old ones.

foodsaving/users/models.py
```diff
@@ -149,9 +141,9 @@ def _send_mail_change_notification(self):
     def _send_welcome_mail(self):
         self._unverify_mail()
 
-        url = '{hostname}/#/verify-mail?key={key}'.format(
+        url = '{hostname}/#/verify-mail?code={code}'.format(
```
tiltec: Ah, the rename from `key` to `code` will break the frontend. Would you rather change the frontend or change it back to `key={code}`?

Also, we could also use the `furl` library to construct the URL (similar to the invitation mail). But not necessary...

foodsaving/users/models.py
```diff
@@ -172,9 +164,9 @@ def _send_welcome_mail(self):
     def send_new_verification_code(self):
         self._unverify_mail()
 
-        url = '{hostname}/#/verify-mail?key={key}'.format(
+        url = '{hostname}/#/verify-mail?code={code}'.format(
```
tiltec: Same here about frontend break.

foodsaving/userauth/serializers.py
```diff
@@ -41,7 +42,7 @@ class Meta:
         model = get_user_model()
         fields = ['id', 'display_name', 'email', 'unverified_email', 'password',
                   'address', 'latitude', 'longitude', 'description', 'mail_verified',
-                  'key_expires_at', 'current_group', 'language', 'photo', 'photo_urls']
+                  'current_group', 'language', 'photo', 'photo_urls']
         read_only_fields = ('unverified_email', 'key_expires_at', 'mail_verified')
```
tiltec: Remove `key_expires_at` from `read_only_fields` too

foodsaving/users/migrations/0017_auto_20171207_1210.py
```diff
@@ -16,6 +16,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='user',
             name='current_group',
-            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to='groups.Group'),
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL,
```
tiltec: unrelated change

foodsaving/groups/migrations/0017_auto_20171207_1210.py
```diff
@@ -16,6 +16,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='group',
             name='active_agreement',
-            field=models.OneToOneField(null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='active_group', to='groups.Agreement'),
+            field=models.OneToOneField(null=True, on_delete=django.db.models.deletion.SET_NULL,
```
tiltec: Unrelated change, too

foodsaving/userauth/api.py
```diff
@@ -100,7 +98,7 @@ def post(self, request):
         """
         Send token to verify e-mail
 
-        requires "key" parameter
+        requires "code" parameter
```
tiltec: Serializer still expects the `key` parameter, I would probably leave it that way for now.

foodsaving/userauth/models.py
```diff
@@ -0,0 +1,57 @@
+from datetime import timedelta
+
+from django.db.models import CharField, ForeignKey, Manager
+from django.db import models
+from django.conf import settings
+from django.utils import crypto, timezone
+
+from foodsaving.base.base_models import BaseModel
+
+# TODO: Ensure each user can have at most one verification code of a certain type at a time.
```
tiltec: Seems good, you can do that with a one-liner by adding a `unique_together` constraint. I used it in the invitations code, for example.

foodsaving/users/models.py
mr-kojo: Oh... sorry for that!
I actually wanted to test everything with my local frontend, then I would have noted that myself. But then I forgot...

mr-kojo:
Successfully tested using my local frontend (should have done that right away, sorry for that)

foodsaving/conversations/migrations/0007_auto_20171207_1210.py
```diff
@@ -16,6 +16,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='conversationparticipant',
             name='seen_up_to',
-            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='conversations.ConversationMessage'),
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL,
```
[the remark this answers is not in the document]
tiltec: Yes, migrations are important! On the other hand, they have been autogenerated in 99% of the cases and the Django migrations generator didn't respect PEP8 out of the box (I didn't try further), so I just said that PEP8 wouldn't be checked there.
Another point is that migrations are usually "throw away code": after they ran successfully, they can be [squashed](https://docs.djangoproject.com/en/2.0/topics/migrations/#migration-squashing) and removed. Edits should _never_ be necessary after they've been merged to master.

Before we start enforcing PEP8 on migrations, I would strongly like to have a `makemigrations` command that produces compliant files. Django won't do it out of the box, see [this bug](https://code.djangoproject.com/ticket/24275), which has been closed as WONTFIX. The solution could be to override the `makemigrations` to run something like `autopep8` automatically.

You are fine to write your custom migrations in PEP8, but I consider style changes to old migrations unnecessary noise (which could have been avoided by having all files in PEP8 from day one)

tiltec:
Cool, looks even better now! 

If you still want to revert the unrelated changes, I'd be very happy. I'm not strongly bothered by it, but would say that it's easier to review a PR which contains only changes related to a single issue.

PR #473, opened 2018-01-26
Add update timestamp to conversation serializer

tiltec:
As of #471, the client receives updates of the conversation status via websocket push. The client might receive the updates out of order, resulting in out-of-date state. 

To prevent these bugs, this PR adds an `updated_at` field to the conversation serializer.

The value of `updated_at` is dependent on the requesting user, because it also takes the update timestamp of `ConversationParticipant` into account (more recent date gets sent). Two signal receivers make sure that `conversation.updated_at` gets changed when a participant gets added or removed.


tiltec:
Frontend support comes with https://github.com/yunity/karrot-frontend/pull/848

tiltec:
The cordova app needs to actively refresh the conversation metadata after it wakes up from background. (Probably best to refresh everything).

PR #476, opened 2018-01-31
Include domain with photo urls sent via socket

nicksellen:
Makes the websocket receiver stuff include the full domain when sending photo urls back, to make it consistent with REST API.

See https://github.com/yunity/karrot-frontend/issues/857.


foodsaving/userauth/tests/test_userauth_api.py
```diff
@@ -162,6 +162,9 @@ def test_upload_and_delete_photo(self):
         response = self.client.get(self.url)
         self.assertTrue('full_size' in response.data['photo_urls'])
         self.assertTrue('thumbnail' in response.data['photo_urls'])
+        print('hostname', settings.HOSTNAME)
```
tiltec: remove print statements

PR #485, opened 2018-02-22
add mobile number to user model

DevipriyaSarkar:
add mobile number to user model, user serializer, userauth serializer and add generation of fake mobile number while creating sample data (referencing https://github.com/yunity/karrot-frontend/issues/552)


DevipriyaSarkar:
Sure. I'll add the migration file too. :)

PR #501, opened 2018-02-27
Rework auth API

### LINKED ISSUE #275: Introduce token-based actions

mr-kojo:
**Steps:**
- [x] E-mail verification
- [x] E-mail change
- [x] Password change
- [x] Password reset
- [x] Account deletion

**Changed API endpoints:**

1. `api/auth/verify_mail/`: 
    - Renamed to `api/auth/email/verify/`
    - Parameter `key` is now called `code`
    - No authentication required anymore
2. `api/auth/resend_verification/`: 
    - Renamed to `api/auth/email/resend_verification_code/`
3. `api/auth/reset_password/`:
    - Renamed to `api/auth/password/request_reset/`
4. `api/auth/change_password/`:
    - Renamed to `api/auth/password/`:
    - Keeps the user logged in (i.e., updates the session key. See [`update_session_auth_hash`](https://docs.djangoproject.com/en/2.0/topics/auth/default/#django.contrib.auth.update_session_auth_hash))
5. `api/auth/user/`:
    - No longer allows to change email address
    - HTTP DELETE expects parameter `code` now

**New API endpoints:**

1. `api/auth/password/reset/`:
    - Parameter `code`: verification code
    - Parameter `new_password`: The new password to set
2. `api/auth/email/`:
    - Parameter `password`: The current password. (Without password confirmation, the new two-factor-authentication would not make much sense.)
    - Parameter `new_email`: The email address to set
    - Allows "undoing" an email change. I.e., if the new email address has not been validated yet, one can switch back to the old one without having to validate it again.
3. `api/auth/user/request_delete/`:
    - No parameters
    - Requires login

Closes #275


mr-kojo:
Closes #275

nicksellen:
Great work! Quite a beast to tackle I think, looks like a lot more rational and structured approach.

> api/auth/reset_password/ - Login is disabled until new password is set.

Seems like I can disable random peoples accounts if I know their email address?

Then, a few smaller/style things:
- there's a few places where `mail` is used instead of `email`, should be consistent
- the endpoints deviate quite a bit from RESTful, it's tricky with this kind of "action" methods, but I could see them being a bit more resource-like, e.g. `auth/email/reset/` or `auth/password/request_reset`, so something like `auth/<resource>/<action>`, and can DELETE the `auth/user` for the final confirmation - _rationale_: 1. making it clearer for someone to get an understand of which resource is likely to be modified, and how so, 2. consistency

... I didn't really finish reviewing, quite a lot to read, but if we're going to tackle the frontend next, then I think that'll be a good chance to try it out too. I assume this can't be merged until the frontend is ready as it will break everything?

Also, HTML emails would be nice :) - can go through it tomorrow.

foodsaving/users/models.py
```diff
@@ -134,21 +143,69 @@ def _send_welcome_mail(self):
...
                 hostname=settings.HOSTNAME,
                 code=VerificationCode.objects.get(user=self, type=VerificationCode.EMAIL_VERIFICATION).code
             )
         }, to=self.unverified_email).send()
 
+    @transaction.atomic
+    def send_account_deletion_verification_code(self):
+        VerificationCode.objects.filter(user=self, type=VerificationCode.ACCOUNT_DELETE).delete()
+        prepare_email('accountdelete_request', self, {
+            'url': '{hostname}/#/delete_account?code={code}'.format(
```
tiltec: At some point, I'd like to put all frontend urls into a file that makes it possible to reuse them (well, at least the ones that we use in the backend). Might also help if we change something about the frontend routing in future.

foodsaving/users/models.py
```diff
@@ -134,21 +143,69 @@ def _send_welcome_mail(self):
...
+                hostname=settings.HOSTNAME,
+                code=VerificationCode.objects.create(user=self, type=VerificationCode.ACCOUNT_DELETE).code
+            )
+        }).send()
+
     @transaction.atomic
     def reset_password(self):
-        new_password = User.objects.make_random_password(length=20)
+        self.set_unusable_password()
```
tiltec: Would removing this line resolve Nicks comment about being able to temporarily disable some else's account by knowing their e-mail address?

tiltec:
Thanks for all your work on this, future users will be grateful to you!

> there's a few places where mail is used instead of email, should be consistent

I agree, that was bad style I introduced. It will require more frontend changes though, I wonder if it makes sense to do it right now. Maybe it does...

nicksellen:
After chatting with @mr-kojo more concrete explanation of what I was meaning with the (slightly) more RESTful URLs:

```
post /api/auth/
-

post /api/auth/change_mail/
put /api/auth/email/


post /api/auth/change_password/
put /api/auth/password/


post /api/auth/logout/
-


post /api/auth/reset_password/
post /api/auth/password/reset/


post /api/auth/reset_password/request/
post /api/auth/password/request_reset/


post /api/auth/token/
-


get /api/auth/user/
-


patch /api/auth/user/
-


post /api/auth/user/
-


post /api/auth/user/delete/
delete /api/auth/user/?code=foo


post /api/auth/user/delete/request/
post /api/auth/user/request_delete/


post /api/auth/verify_mail/
post /api/auth/email/verify/


post /api/auth/verify_mail/resend_verification_code/
post /api/auth/email/resend_verification_code/
```

foodsaving/users/models.py
mr-kojo: Yes, done via a8836ab

foodsaving/users/models.py
```diff
@@ -138,21 +147,67 @@ def _send_welcome_mail(self):
...
+    @transaction.atomic
+    def change_password(self, new_password):
         self.set_password(new_password)
         self.save()
-        prepare_email('newpassword', self, {'password': new_password}).send()
+        prepare_email('passwordchange', self, {}).send()
+        VerificationCode.objects.filter(user=self, type=VerificationCode.PASSWORD_RESET).delete()
+
+    @transaction.atomic
+    def delete_(self):
```
tiltec: Maybe call it `erase()`?

tiltec:
Done merging #518 

I made a few more related changes:
- move all `prepare` methods into `foodsaving/users/emails.py`
- move all frontend urls into `utils/frontend_urls.py`
- rename email methods and templates to follow the `*_request`, `*_success` scheme

It also contains unrelated changes:
- message author in conversation message notifications is now clickable
- added conversation messages to sample data

foodsaving/utils/tests/test_email_utils.py
```diff
@@ -22,19 +24,19 @@ def test_emailinvitation(self):
...
             invited_by=self.user
         )
-        email = email_utils.prepare_emailinvitation_email(invitation)
-        self.assertEqual(len(email.alternatives), 0)
+        email = foodsaving.invitations.emails.prepare_emailinvitation_email(invitation)
+        self.assertEqual(len(email.alternatives), 1)
         self.assertEqual(email.to[0], 'bla@bla.com')
         self.assertIn(self.group.name, email.body)
         self.assertIn(self.user.display_name, email.body)
         self.assertIn(str(invitation.token), email.body)
```
mr-kojo: I observed a funny bug:
MJML inserted a newline (`\n`) in the middle of the invitation link:

```code
======================================================================
FAIL: test_emailinvitation (foodsaving.utils.tests.test_email_utils.TestEmailUtils)
  File "/usr/lib/python3.5/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.5/unittest/case.py", line 601, in run
    testMethod()
  File "/home/lifejam/workspace/karrot/karrot-backend/foodsaving/utils/tests/test_email_utils.py", line 31, in test_emailinvitation
    self.assertIn(str(invitation.token), email.body)
  File "/usr/lib/python3.5/unittest/case.py", line 1080, in assertIn
    self.fail(self._formatMessage(msg, standardMsg))
  File "/usr/lib/python3.5/unittest/case.py", line 666, in fail
    raise self.failureException(msg)
AssertionError: '7a4d2eff-ae15-47c7-acf6-33d16c5ef9fe' not found in '# You\'re invited to join Steven Collins!  \n  \nTyler Cruz invited you to join the "Steven Collins" community on karrot.world!\n\nIf you\'d like to join, please click the link below:  \n  \n[Follow the invitation][1]  \n  \n[][2]  \n  \n[Join our Facebook Group!][3] • made with love by [foodsaving worldwide][4]\n\n   [1]: http://localhost:8080/#/signup?invite=7a4d2eff-\nae15-47c7-acf6-33d16c5ef9fe&email=bla@bla.com\n\n   [2]: http://localhost:8080\n\n   [3]: https://www.facebook.com/groups/foodsaving.worldwide/\n\n   [4]: https://foodsaving.world\n\n'

```

tiltec: That's probably the `html2text` library. If you switch the test to assert in HTML directly, they should be more reliable.

mr-kojo: That's right, `html2text` was causing the problem. It's fixed now.

foodsaving/utils/frontend_urls.py
```diff
@@ -46,7 +46,7 @@ def invite_url(invitation):
 
 
 def user_delete_url(code):
-    return '{hostname}/#/user/delete?code={code}'.format(
+    return '{hostname}/#/delete-user?code={code}'.format(
```
mr-kojo: `user/delete` is in conflict with `user/:id`. To resolve this, I can think of 3 options:
1. Change `user/delete` to some other URL (this is what I've done)
2. Change `user/:id` to some other URL
3. Add `user/delete` to the vue-router before `user/:id`. This would be the best option in my opinion, but it seems to be difficult considering the implementation of `router.js` (which loads main-layout URLs before splash-layout URLs).

tiltec: 1. seems easiest to me
2. would break existing URLs
3. is a bit tricky, maybe it would make sense to rethink our route implementation. For example, `groups/create` and `groups/:id` both exist and mostly work fine, but I also had a case where it failed in a subtle way. Would not recommend right now.

mr-kojo:
Ready to be merged together with https://github.com/yunity/karrot-frontend/pull/934

PR #503, opened 2018-02-28
WIP: Added new 'mark_user_active' endpoint.

djembejohn:
Added a new 'mark_user_active' endpoint to groups for the frontend to report that a user has been active in a group. Added a new last_seen field to the GroupMembership model which is updated when the endpoint is called.


foodsaving/groups/models.py
```diff
@@ -96,6 +96,7 @@ class GroupMembership(BaseModel):
     group = models.ForeignKey(Group, on_delete=models.CASCADE)
     user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
     roles = ArrayField(TextField(), default=list)
+    last_seen = DateTimeField(default=timezone.now)
```
nicksellen: I would suggest to name this `lastseen_at` so it  is consistent with `ChannelSubscription.lastseen_at` field (and other `*_at` fields containing DateTimes)

PR #507, opened 2018-03-01
Emoji reactions to messages

mrkvon:
Add and remove emoji reactions to messages.

- [x] `POST /messages/{id}/reactions/`
- [x] `DELETE /messages/{id}/reactions/{name}/`
- [x] Include list of reactions when retrieving messages
- [x] More tests, validation, etc.
- [ ] Refactor to use django-rest-framework more efficiently
- [ ] Websocket


foodsaving/conversations/api.py
```diff
@@ -101,6 +106,35 @@ class ConversationMessageViewSet(
...
+        filter_fields=('name',)
+    )
+    def reactions(self, request, pk):
+        """route for POST /messages/{id}/reactions/ with body {\"name\":\"emoji_name\"}"""
+
+        instance = get_object_or_404(ConversationMessage.objects, id=pk)
+        reaction = ConversationMessageReaction(message=instance, name=request.data['name'], user=request.user)
+        try:
+            reaction.save()
+        except IntegrityError:
```
tiltec: I guess that happens when a user adds the same emoji two times?

mrkvon: Indeed it does. It is changed in the last commit though.

foodsaving/conversations/helpers.py
```diff
@@ -0,0 +1,29 @@
+import json
+
+# load list of lists of alternative names of emojis
+# gathered here:
+# https://github.com/markdown-it/markdown-it-emoji/ (continue)
+#    blob/f1bc7217e9eb4d954c145faf316718e263778e34/lib/data/full.json
+with open('foodsaving/conversations/emoji_names.json') as json_data:
```
tiltec: Quite a list! Maybe you could use the library from here: https://github.com/facelessuser/pymdown-extensions/blob/master/pymdownx/twemoji_db.py ?
We already use it for rendering emojis in message notifications, so you could directly import it via `from pymdownx import twemoji_db`.

Unrelated thought: the frontend to have a list of all emoji names, which makes me think if we should switch to a reduced version: https://github.com/markdown-it/markdown-it-emoji/blob/master/lib/data/light.json

foodsaving/conversations/api.py
```diff
@@ -41,6 +46,27 @@ def has_permission(self, request, view):
         return True
 
 
+class IsMessageConversationParticipant(BasePermission):
+    "Is the specified message in a conversation which user is member of?"
+
+    message = _('You are not in the conversation')
+
+    def has_permission(self, request, view):
```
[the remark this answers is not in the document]
tiltec: I'd be happy to help you with this, my IDE works again :)

mrkvon:
@tiltec the issues from your code review should now be fixed.

yunity/karrot-frontend#926 still awaits fixing.

foodsaving/conversations/api.py
```diff
@@ -101,6 +115,52 @@ class ConversationMessageViewSet(
...
+        }
+
+        serializer = ConversationMessageReactionSerializer(data=data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+    @detail_route(
+        methods=('DELETE',),
+        url_path='reactions/(?P<name>[a-z0-9_+-]+)',
```
tiltec: Can you add the trailing slash here, as mentioned in https://github.com/yunity/karrot-frontend/pull/926#discussion_r172047693 ?

foodsaving/conversations/api.py
```diff
@@ -101,6 +115,52 @@ class ConversationMessageViewSet(
...
+        url_name='remove_reaction',
+        permission_classes=(IsAuthenticated, IsMessageConversationParticipant)
+    )
+    def remove_reaction(self, request, pk, name):
+        """route for DELETE /messages/{id}/reactions/{name}/"""
+
+        name = EmojiField.to_internal_value(None, name)
+
+        message = get_object_or_404(ConversationMessage.objects, id=pk)
```
tiltec: Optional: according to https://docs.djangoproject.com/en/2.0/_modules/django/shortcuts/#get_object_or_404 you can also pass a model to `get_object_or_404`. Makes the code a bit shorter: `message = get_object_or_404(ConversationMessage, id=pk)` (especially in the second case below)

tiltec:
Thanks @mrkvon! Only add the trailing slash for the remove endpoint (I mentioned it in the frontend review, didn't notice it here first), then I'll merge :)

foodsaving/conversations/api.py
mrkvon: I overlooked this.
However when I try to add trailing slash here, the tests (which contain the trailing slash) start failing. When I put 2 trailing slashes into tests, they pass again.
So I believe the trailing slash is taken care of on a different level.
Just observation. Don't know why that happens. Perhaps [this](http://www.django-rest-framework.org/api-guide/routers/#defaultrouter)?

tiltec:
Ah nice, I didn't expect that. So it seems fine. And the frontend without the trailing slash probably worked because it was redirected.

PR #508, opened 2018-03-01
Report aggregate activity

nicksellen:


foodsaving/groups/tasks.py
```diff
@@ -0,0 +1,12 @@
+from huey import crontab
+from huey.contrib.djhuey import db_periodic_task
+
+from foodsaving.groups import stats
+from foodsaving.groups.models import Group
+
+
+@db_periodic_task(crontab(minute='*'))
```
nicksellen: Oh, I left it to report every minute... was just going to make sure it is all working, then push a change to reduce it to, perhaps, once per day.

foodsaving/groups/stats.py
```diff
@@ -0,0 +1,72 @@
...
+    write_points([{
+        'measurement': 'karrot.group.members',
+        'tags': {
+            'group': str(group.id)
+        },
+        'fields': {
+            'count_total': group.members.count(),
+            'count_active_1d': members_active_within(days=1),
+            'count_active_7d': members_active_within(days=7),
+            'count_active_30d': members_active_within(days=30),
```
nicksellen: Maybe a longer time period would be useful too. 60d? 90d? 1 year...

PR #509, opened 2018-03-02
Mark inactive users and send them an email

djembejohn:
I have implemented the inactive users script.

Added an "isactive" field to GroupMembership to indicate is a user is considered active in a group or not. Added a script process_inactive_users.py to process inactive users to commands. Added English email templates for the script to send emails.


nicksellen:
Great to have this feature!

Although I don't see the need to add an `inactive` field - as it contains no more information than the `lastseen_at` field + the config options for inactivity thresholds, so can be derived whenever needed.

A slight difference is that if we have the `inactive` field, we can change the thresholds without affecting past decisions about inactivity made with previous threshold values. I'm not sure if this is a positive thing or not.

For returning active status in the API, it's easy to return a derived value with a serializer method field.

In general I think it's a good idea to minimise the number of places where state is stored or modified.

Also, would be nice to have some tests, but maybe pending outcome of thinking about the above :)

Also also, I think we can move away from using commands for periodic tasks, as we have the Huey task runner which can do this for us. Using this approach you have a `tasks.py` inside each django app that have functions annotated with `@periodic_task` or `@db_periodic_task` with some crontab parameters.

tiltec:
@nicksellen Don't we need a field to store if we already sent out the e-mail to the user that became inactive?

djembejohn:
Yes, I agree with @tiltec. There are too many things that could go wrong with the script otherwise.

nicksellen:
Yes, good point. I did not consider that properly.

Although I would enjoy more consistent between the naming of boolean fields, we now have:

1. active (in groups) - 7 cases fit this pattern (no `is_` prefix)
2. is_date_changed (in pickups) - 6 cases fit this pattern (with `is_` prefix)
3. isactive (in group memberships in this branch) - nothing else uses this pattern

As both 1 and 2 are in use in quite a few places, and 1 is most used, I'd propose to rename it to `active`.

@tiltec pointed out though that it is a bit confusing having the builtin django field `is_active` (on the user) at the same time as our own `active` field (on the group membership) - I would imagine someone will get confused at some point, but it doesn't seem a big problem to me right now.

It does touch on a wider topic as really the field is about storing the fact that an email notification was sent, and there might be lots of other cases that arise where we want to record information like this, instead of littering the database with boolean fields with names that are not obvious that they are really just about recording a notification. I think this topic is out of scope for this particular issue though.

So, incase it's not clear from my rambling - I propose to rename this field from `isactive` to `active`.

foodsaving/groups/tasks.py
```diff
@@ -15,3 +22,38 @@ def record_stats():
...
+    print('Sending email to userid=', user.id, 'regarding removal from groupid=', group.id)
+    email = prepare_user_removed_from_group_email(user, group)
+    email.send()
+
+        
+@db_periodic_task(crontab(hour='2',minute='0'))
+def process_inactive_users ():
+    now = timezone.now()
+
+    print("Processing inactive users")
```
tiltec: I'd remove those print statements, I guess they will clutter the log files otherwise.

foodsaving/groups/tests/test_tasks.py
```diff
@@ -0,0 +1,47 @@
...
+from foodsaving.groups.factories import GroupFactory
+from foodsaving.groups.models import Group, GroupMembership
+from foodsaving.groups.tasks import process_inactive_users
+from foodsaving.users.factories import UserFactory
+from config import settings
+
+from datetime import timedelta
+
+
+# This probably shouldn't be APITestCase???
```
tiltec: It doesn't really matter, APITestCase just provides an additional client. You could make it `TestCase` though to make clear that you're not using any client here.
It can be imported from `from django.test import TestCase`.

foodsaving/groups/tests/test_tasks.py
```diff
@@ -0,0 +1,47 @@
...
+        self.membership2 = GroupMembership.objects.get(group=self.group, user=self.member2)
+        self.membership2.lastseen_at = remove_email_date
+        self.membership2.active = False
+        self.membership2.save()
+
+        mail.outbox = []
+
+    def test_process_user_inactive(self):
+        process_inactive_users()
+        self.assertEqual(len(mail.outbox), 2)
```
tiltec: You should probably check if the user got the right e-mail. You can access the subject of an email with `mail.outbox[0].subject` and compare it to a string. For example:

```
member1_email = next(email for email in self.outbox if email.to[0] == self.member1.email)
self.assertIn('Please come back to', member1.subject)
```

I should probably write a helper to assist finding the email to a user...

foodsaving/groups/tests/test_tasks.py
```diff
@@ -0,0 +1,47 @@
...
+from foodsaving.users.factories import UserFactory
+from config import settings
+
+from datetime import timedelta
+
+
+# This probably shouldn't be APITestCase???
+class TestProcessInactiveUsers (APITestCase):
+    def setUp(self):        
+        self.admin = UserFactory()  # has membership management rights
```
tiltec: Comment seems wrong. Calling this user `self.active_user` seems more appropriate.

djembejohn:
I think something should be logged though? Otherwise we won't be able to tell what's been happening when something goes wrong.


On 3 March 2018 13:17:46 CET, Tilmann Becker <notifications@github.com> wrote:
>tiltec commented on this pull request.
>
>
>
>> +    print('Sending email to userid=', user.id, 'regarding inactivity
>in groupid=', group.id)
>+    email = prepare_user_inactive_in_group_email(user, group)
>+    email.send()
>+
>+
>+def send_removal_from_group_notification_to_user(user, group):
>+    print('Sending email to userid=', user.id, 'regarding removal from
>groupid=', group.id)
>+    email = prepare_user_removed_from_group_email(user, group)
>+    email.send()
>+
>+        
>+@db_periodic_task(crontab(hour='2',minute='0'))
>+def process_inactive_users ():
>+    now = timezone.now()
>+
>+    print("Processing inactive users")
>
>I'd remove those print statements, I guess they will clutter the log
>files otherwise.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/yunity/karrot-backend/pull/509#pullrequestreview-100973531

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

foodsaving/groups/tasks.py
```diff
@@ -15,3 +22,38 @@ def record_stats():
...
+
+    remove_threshold_date = now - timedelta(days=settings.NUMBER_OF_DAYS_UNTIL_REMOVED_FROM_GROUP)
+    print("Removing inactive users in groups who have been inactive since:",
+          datetime.strftime(remove_threshold_date, '%Y-%m-%d'))
+    for gm in GroupMembership.objects.all().filter(lastseen_at__lte=remove_threshold_date, active=False):
```
tiltec: Calling the loop variable `gm` seems a bit cryptic. `membership` might be more readable.

foodsaving/groups/tests/test_tasks.py
```diff
@@ -0,0 +1,47 @@
...
+        self.membership2.save()
+
+        mail.outbox = []
+
+    def test_process_user_inactive(self):
+        process_inactive_users()
+        self.assertEqual(len(mail.outbox), 2)
+        self.membership1.refresh_from_db()
+        self.assertEqual(self.membership1.active, False)
+        self.assertEqual(len(GroupMembership.objects.all()), 2)
```
tiltec: Checking if the right user has been removed could be good.

foodsaving/utils/email_utils.py
```diff
@@ -92,107 +92,6 @@ def prepare_emailinvitation_email(invitation):
     }, to=invitation.email)
 
 
-def calculate_group_summary_dates(group):
```
tiltec: I wonder why this shows up as deleted?

foodsaving/groups/tests/test_tasks.py
djembejohn: Hmmm... I'm not sure that user needs to exist actually. I'm going to remove it.

foodsaving/groups/tests/test_tasks.py
djembejohn: Checking the content of emails is reasonable. The contents of the emails should probably be agreed first though.

foodsaving/groups/tests/test_tasks.py
djembejohn: No problem

foodsaving/utils/email_utils.py
djembejohn: I did a merge with master, I maybe did the wrong thing. Would be good to check I didn't.

foodsaving/groups/tests/test_tasks.py
djembejohn: Done

foodsaving/groups/tests/test_tasks.py
tiltec: Hang on, it makes sense to keep it. Just to make sure that an active user doesn't get these e-mails :)

tiltec:
> I think something should be logged though? Otherwise we won't be able to tell what's been happening when something goes wrong.

From my POV, I expect the application to throw an exception if something goes wrong. But maybe it makes sense to keep the print statements as long as we're not in production. I'll talk with Nick about it, as he also introduced print statements in his tasks.

foodsaving/groups/tests/test_tasks.py
djembejohn: Have put it back, it now has separate tests for the three different user types

foodsaving/groups/templates/user_inactive_in_group.subject.jinja2
```diff
@@ -0,0 +1 @@
+[{{ site_name }}] {% trans %}We're missing you!{% endtrans %}
```
tiltec: I guess the subject should include the group name, something like "{{ group_name }} is missing you!"

foodsaving/groups/templates/user_inactive_in_group.mjml
```diff
@@ -0,0 +1,33 @@
...
+                <mj-column>
+                    <mj-text>
+                        <p>
+			{% trans %}Hey, we're missing you!{% endtrans %}
+		    	</p>
+                        <p>
+                            {% trans %}It looks like you've not visited the <a href="{{ group_url }}">{{ group_name }}</a> group for over {{ num_days_inactive }} days.{% endtrans %}
+			</p>
+                        <p>
+                            {% trans %}If you don't visit the group in the next {{ inactivity_deletion_days }} we will sadly have to remove you from that group. {% endtrans %}
```
tiltec: I'd put this paragraph at the end of the e-mail (below the button). First lure them, then threat them ;)
Maybe also give a reason:

```
If you are not interested in {{ group_name }} anymore, you don't need to click. After {{ inactivity_deletion_days }}, you will get removed from {{ group_name }} to keep the member list up to date.
```

Generally, I'd mention the group_name in the e-mail more often instead of referring to it as "the group".

foodsaving/groups/templates/user_removed_from_group.mjml
```diff
@@ -0,0 +1,33 @@
...
+        <mj-container>
+            <mj-include path="../../base/templates/_header.mjml"/>
+            <mj-section>
+                <mj-column>
+                    <mj-text>
+                        <p>
+			{% trans %}Hey,{% endtrans %}
+		    	</p>
+                        <p>
+                            {% trans %}It has been over {{ inactivity_deletion_days }} days since you visited the "{{ group_name }}" group.{% endtrans %}
```
tiltec: "...visited {{ group_name }}"

foodsaving/groups/templates/user_removed_from_group.subject.jinja2
```diff
@@ -0,0 +1 @@
+[{{ site_name }}] {% trans %}We missed you{% endtrans %}
```
tiltec: How about "You have been removed from {{ group_name }}" ?

foodsaving/groups/tasks.py
```diff
@@ -22,6 +29,34 @@ def record_group_stats():
...
+
+    remove_threshold_date = now - timedelta(days=settings.NUMBER_OF_DAYS_UNTIL_REMOVED_FROM_GROUP)
+    for gm in GroupMembership.objects.all().filter(lastseen_at__lte=remove_threshold_date, active=False):
+        email = prepare_user_removed_from_group_email(gm.user, gm.group)
+        email.send()
+        gm.delete()
+        count_users_removed += 1
+
+    inactive_threshold_date = now - timedelta(days=settings.NUMBER_OF_DAYS_UNTIL_INACTIVE_IN_GROUP)
+    for gm in GroupMembership.objects.all().filter(lastseen_at__lte=inactive_threshold_date, active=True):
```
tiltec: Ah, you don't actually need the call to `all()` here and above. Makes the line a bit shorter.

foodsaving/groups/stats.py
```diff
@@ -44,11 +44,11 @@ def group_summary_email(group, recipient_count):
     }])
 
 
-def periodic_task(name):
+def periodic_task(name, extra_fields=dict):
```
tiltec: That's not how python default values work. extra_field will be a callable, not an actual dict.

The standard way out of this is to set `None` and set the actual default in the function: https://reinout.vanrees.org/weblog/2012/04/18/default-parameters.html (solution is at the bottom)

foodsaving/utils/email_utils.py
```diff
@@ -92,6 +99,107 @@ def prepare_emailinvitation_email(invitation):
     }, to=invitation.email)
 
 
+def calculate_group_summary_dates(group):
```
tiltec: I thought Nick moved that code to `groups/emails.py`? Seems like a merge problem.

foodsaving/groups/stats.py
nicksellen: In anticipation of a review from @tiltec, it's my fault that I totally lied to you about how python default value work ;)

foodsaving/groups/tests/test_tasks.py
```diff
@@ -0,0 +1,53 @@
...
+from foodsaving.groups.tasks import process_inactive_users
+from foodsaving.users.factories import UserFactory
+from config import settings
+
+from datetime import timedelta
+
+
+class TestProcessInactiveUsers (TestCase):
+    def setUp(self):
+        self.activeUser = UserFactory()
```
tiltec: Please use snake_case for variables, as per [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names). I forgot to add this to the code style checker, will do soon.

foodsaving/groups/templates/user_inactive_in_group.mjml
djembejohn: I put it to the following. Having the group name twice in two sentences didn't seem good style.

If you are not interested in "{{ group_name }}" anymore, you don't need to click. After {{ inactivity_deletion_days }}, you will be removed from the group to keep the member list up to date.

foodsaving/groups/templates/user_removed_from_group.subject.jinja2
```diff
@@ -0,0 +1 @@
+[{{ group_name }}] {% trans %}Group removal{% endtrans %}
```
tiltec: "Group removal" sounds like the group is getting removed... how about "Removal from group"?

nicksellen:
I thought of two scenarios where long term inactive users will be removed without the grace period:
1. if during the migration for adding the lastseen_at, we set it to a more realistic time (instead of now())
2. if the script is not run for a while

I would like to implement 1 (seperately), so that when we deploy to production we have some slightly more useful data about user activity immediately, instead of having to wait for 7d, 30d, 60d, 90d for those periods to become meaningful.

Perhaps 2 seems unlikely in normal operation, but there could be cases, e.g. someone shuts down their karrot instance for some time, then starts it up again, within two executations of the task all the users will be removed from all the groups.

I think if it was a non-destructive operation then I wouldn't be so concerned about that kind of thing, but removing users from groups could be a big deal for someone, so seems worth getting it right.

Thinking about it, it makes sense for the removal to happen _n_ days after the point they were made inactive.

So, instead of the current behaviour:
```
|-------- >30 days ---------| make user inactive
|-------------------------- >90 days ----------------------------| remove user from group
```

It would be:
```
|------- >30 days ----------| make user inactive
                                     |------------- >60 days ------------| remove user from group
```

To do this the `active` field could be removed, and an `inactive_at` DateTime field added. Filtering for active users would be achieved with `inactive_at is null`.

Chatting a bit with @tiltec just now it seems to make sense to only implement inactive marking for now, as that would already offer very useful functionality immediately, then leave the implementation of full deletion until the future.

Sorry to add more thoughts at this point, but it sometimes take a bit of time to think through the scenarios.

djembejohn:
I have added an inactive_at field which indicates the user has been sent an email.

Thinking about removing users. I have currently commented out that code and there is no test now.

I wondered about adding a 'removed' field. However this is probably more trouble that it's worth as it makes dealing with the group membership table very tricky as you have to remember to filter out removed users every time you query it.

It should probably add a history element that user x was deleted from group y. Then it would be possible to undo this from the history if necessary.

It's unlikely I'll find time to work on this for a while now.

nicksellen:
Should be good to go! Maybe a review from @tiltec ...

foodsaving/groups/templates/user_inactive_in_group.subject.jinja2
```diff
@@ -0,0 +1 @@
+[{{ group_name }}] {% trans %}is missing you!{% endtrans %}
```
tiltec: Still not sure about the subject, it seems a bit weird to have brackets in a sentence

nicksellen: Agree

nicksellen: ... and updated

PR #516, opened 2018-03-03
Summary emails ignore some emails, add task

nicksellen:
This should start actually sending the summary emails now.


foodsaving/groups/stats.py
```diff
@@ -34,6 +34,24 @@ def group_activity(group):
...
+    write_points([{
+        'measurement': 'karrot.email.group_summary',
+        'tags': {
+            'group': str(group.id)
+        },
+        'fields': {'value': 1, 'recipient_count': recipient_count},
+    }])
+
+
+def periodic_task(name):
```
tiltec: Naming seems quite generic, but fine for now

foodsaving/groups/tasks.py
```diff
@@ -1,17 +1,50 @@
...
+    stats.periodic_task('group__send_summary_emails')
+
+    groups = Group.objects.annotate(member_count=Count('members')).filter(member_count__gt=0)
+    for group in groups:
+
+        from_date, to_date = emails.calculate_group_summary_dates(group)
+
+        if not group.sent_summary_up_to or group.sent_summary_up_to < to_date:
+
+            print('sending summary for [{}] {} to {}'.format(group.name, from_date, to_date))
```
tiltec: Ah, I just commented in @djembejohn s PR to remove the print statements. My guess is that they show up in the huey service logs, so I'd remove the to have less clutter. Did you try it out?

foodsaving/groups/tests/test_emails.py
```diff
@@ -9,45 +10,53 @@
...
+            sorted([member.email for member in expected_members])
         )
         self.assertNotIn(self.user_without_notifications.email, emails[0].to)
 
     def test_creates_three_emails_for_three_languages(self):
         n = 5
 
         for _ in list(range(n)):
-            self.group.add_member(UserFactory(language='en'))
+            self.group.add_member(VerifiedUserFactory(language='en', mail_verified=True))
```
tiltec: No need to pass `mail_verified=True`.

foodsaving/webhooks/models.py
```diff
@@ -1,11 +1,25 @@
...
+            event__in=settings.EMAIL_EVENTS_AVOID
+        ).values('address')
+
+
 class EmailEvent(BaseModel):
     id = BigAutoField(primary_key=True)
     address = TextField()
     event = CharField(max_length=255)
     payload = JSONField()
+    objects = EmailEventManager()
```
tiltec: I'd put this at the top of `EmailEvent` and separate it by space, as it's a meta field, not a database field.

foodsaving/groups/stats.py
nicksellen: Yes, I think it would be used across apps, so belongs somewhere more general, but didn't find the place yet...

nicksellen:
Fixed all the things @tiltec

PR #517, opened 2018-03-03
Add approval role to group members

NerdyProjects:
Adds the group join approval possibility to models/API.

It might be complete, but I would like to go another time through all the accesses/add more tests, so consider it WIP.

Might be ready for a "style-review" :-)

### DIFF
--- a/foodsaving/groups/api.py
+++ b/foodsaving/groups/api.py
@@ -102,7 +102,8 @@ class GroupViewSet(
         # can join all groups but only do anything else to groups where user is member of
         if self.action == 'join':
             return self.queryset
-        return self.queryset.filter(groupmembership__user=self.request.user, groupmembership__roles__contains=[APPROVED])
+        return self.queryset.filter(groupmembership__user=self.request.user,
+                                    groupmembership__roles__contains=[APPROVED])
 
     @detail_route(
         methods=['POST'],
--- a/foodsaving/groups/api.py
+++ b/foodsaving/groups/api.py
@@ -102,7 +102,7 @@ class GroupViewSet(
         # can join all groups but only do anything else to groups where user is member of
         if self.action == 'join':
             return self.queryset
-        return self.queryset.filter(members=self.request.user, roles__contains=[APPROVED])
+        return self.queryset.filter(groupmembership__user=self.request.user, groupmembership__roles__contains=[APPROVED])
 
     @detail_route(
         methods=['POST'],
--- a/foodsaving/groups/api.py
+++ b/foodsaving/groups/api.py
@@ -142,10 +142,15 @@ class GroupViewSet(
     )
     def mark_user_active(self, request, pk=None):
         """Mark that the logged-in user is active in the group"""
-        gm = get_object_or_404(GroupMembership.objects, group=pk, user=request.user, roles__contains=[GROUP_APPROVED_MEMBER])
-        gm.lastseen_at = timezone.now()
-        gm.save()
-        stats.group_activity(gm.group)
+        membership = get_object_or_404(
+            GroupMembership,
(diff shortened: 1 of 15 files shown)


foodsaving/groups/models.py
```diff
@@ -50,8 +57,14 @@ def send_notifications(self):
                 ):
                     p.notify_upcoming_via_slack()
 
-    def add_member(self, user, history_payload=None):
+    # Adds a person to the group marked as being an applicant
+    def add_applicant(self, user):
         GroupMembership.objects.create(group=self, user=user)
+        self.create = History.objects.create(typus=HistoryTypus.GROUP_APPLY, group=self, users=[user, ], )
+
+    # Adds a "full" member to the group, e.g. grants the status of a normal member.
```
tiltec: Thanks for adding documentation. If you add them as docstring, the IDE will help you even more: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

foodsaving/history/tests/test_api.py
```diff
@@ -295,6 +295,6 @@ def test_filter_by_date(self):
         self.client.force_login(self.member)
         self.client.post('/api/groups/', {'name': 'xyzabc', 'timezone': 'Europe/Berlin'})
         response = self.get_results(history_url, data={'date_1': timezone.now()})
-        self.assertEqual(len(response.data), 1)
+        self.assertEqual(len(response.data), 2)
```
tiltec: Does this create a history event because the group creator joined the group?

foodsaving/groups/tests/test_model.py
```diff
@@ -23,3 +23,16 @@ def test_roles_initialized(self):
         group = GroupFactory(members=[user])
         membership = GroupMembership.objects.filter(user=user, group=group).first()
         self.assertIn(roles.GROUP_MEMBERSHIP_MANAGER, membership.roles)
+
+    def test_active_member_count(self):
+        user = [UserFactory() for i in range(2)]
```
tiltec: I tend to write `_` if the variable is not used (instead of `i`)

foodsaving/groups/models.py
```diff
@@ -37,6 +38,12 @@ class Group(BaseModel, LocationModel, ConversationMixin):
         on_delete=models.SET_NULL
     )
 
+    def active_member_count(self):
```
tiltec: maybe should be `approved_member_count`? It's a different concept than being `active`.

foodsaving/groups/api.py
```diff
@@ -98,9 +99,10 @@ def partial_update(self, request, *args, **kwargs):
         return super().partial_update(request, *args, **kwargs)
 
     def get_queryset(self):
+        # can join all groups but only do anything else to groups where user is member of
         if self.action == 'join':
             return self.queryset
-        return self.queryset.filter(members=self.request.user)
+        return self.queryset.filter(groupmembership__user=self.request.user, groupmembership__roles__contains=[APPROVED])
```
tiltec: I guess this is one of the bits that needs to be fiddled with in future. Non-approved users should have some access to the groups they applied to, but not full access.

foodsaving/history/tests/test_api.py
NerdyProjects: yep

foodsaving/groups/tests/test_model.py
NerdyProjects: you are right, thanks. My python skills went a bit down...

foodsaving/groups/api.py
NerdyProjects: yep. But as the data in the serializer may also change, I see no reason for doing something there yet.

PR #519, opened 2018-03-03
Group member approval role, try #2

tiltec:
Previous one #517 messed up dev.karrot.world, I guess because it didn't add a data migration to add all existing group members as `approved`.

I didn't try it out properly anyways, so here's the next PR.


nicksellen:
Well, it was more broken than that, as it's a reasonable state to be a non-approved member of a group, but the frontend (and me) had no idea what was going on.

Or is it intended to just make everyone approved and not supported non-approved members for now?

tiltec:
Yes, the goal was to make a backwards-compatible change that adds all group members as `approved`. Then we can merge and carry on with life ;)

tiltec:
As far as I can see, there are two things left:

- [ ] data migration to make all existing members `approved`, including a migration test
- [ ] manual testing with frontend

tiltec:
This needs testing with frontend, otherwise it can get merged.

foodsaving/groups/models.py
```diff
@@ -120,6 +140,12 @@ def add_roles(self, roles):
         for role in roles:
             if role not in self.roles:
                 self.roles.append(role)
+                if role is GROUP_APPROVED_MEMBER:
+                    History.objects.create(
+                        typus=HistoryTypus.GROUP_JOIN,
+                        group=self.group,
+                        users=[self.user, ],
+                    )
```
nicksellen: Hmm, not sure about this:
- we could consider joining the group and being approved to be distinct
- there are probably places that don't go through this add_role method (can it be done via signal instead?)
- I don't see the corresponding group leave history item for remove role

foodsaving/groups/models.py
```diff
@@ -37,6 +38,12 @@ class Group(BaseModel, LocationModel, ConversationMixin):
         on_delete=models.SET_NULL
     )
 
+    def approved_member_count(self):
+        return self.members_with_all_roles([GROUP_APPROVED_MEMBER]).count()
+
+    def members_with_all_roles(self, roles):
+        return self.members.filter(groupmembership__roles__contains=roles)
```
nicksellen: I removed the original `members_with_*`  method, in favour of a more reusable approach, although maybe it's not clear yet how to do that and if it's better.

foodsaving/groups/models.py
NerdyProjects: Hmm, this might need more discussion then, e.g. what is the group join history type meant to do?
I see that I currently got it inconsistent.

foodsaving/groups/receivers.py
```diff
@@ -29,9 +30,13 @@ def group_member_added(sender, instance, **kwargs):
     """When a user is removed from a conversation we will notify them."""
     group = instance.group
     user = instance.user
+    roles = instance.roles
     conversation = Conversation.objects.get_or_create_for_target(group)
-    conversation.join(user)
-    stats.group_joined(group)
+    if GROUP_APPROVED_MEMBER in roles:
+        conversation.join(user)
+        stats.group_joined(group)
```
nicksellen: This will be called whenever an approved group membership is saved (e.g. all the lastseen_at times) so our stats will be wildly optimistic :)

nicksellen: ... although looking at master, this is already the case

nicksellen:
The frontend is still totally dependent on the user having the approval role, and I think it needs quite a bit more careful thinking to work out what to reveal and what not to... not to mention how to consider these non-approved "members" in general...

... but I think this should be mergable now, I did the "click randomly around the frontend" "test" and it passed. Just maybe a review again @tiltec

foodsaving/groups/models.py
```diff
@@ -38,28 +49,41 @@ class Group(BaseModel, LocationModel, ConversationMixin):
...
-        History.objects.create(
-            typus=HistoryTypus.GROUP_LEAVE,
-            group=self,
-            users=[user, ]
-        )
+        if self.is_member(user):
+            History.objects.create(
+                typus=HistoryTypus.GROUP_LEAVE,
+                group=self,
+                users=[user, ]
```
nicksellen: This could be handled in the receiver already

foodsaving/groups/models.py
```diff
@@ -38,28 +49,41 @@ class Group(BaseModel, LocationModel, ConversationMixin):
...
+            )
         GroupMembership.objects.filter(group=self, user=user).delete()
 
     def is_member(self, user):
-        return not user.is_anonymous and GroupMembership.objects.filter(group=self, user=user).exists()
+        return self.is_member_with_role(user, [GROUP_APPROVED_MEMBER])
```
nicksellen: Would prefer to keep this to just database membership check...

foodsaving/groups/models.py
```diff
@@ -97,14 +121,16 @@ def get_default_notification_types():
...
     def with_notification_type(self, type):
         return self.filter(notification_types__contains=[type])
 
+    def with_role(self, role):
+        return self.filter(roles__contains=[role])
+
 
 class GroupMembership(BaseModel):
-    objects = GroupMembershipManager()
+    objects = GroupMembershipQuerySet.as_manager()
```
nicksellen: Look ma! Fancy custom querysets :)

foodsaving/tests/utils.py
```diff
@@ -1,12 +1,15 @@
+from contextlib import contextmanager
+# Mostly based on this nice persons article:
+#   https://www.caktusgroup.com/blog/2016/02/02/writing-unit-tests-django-migrations/
```
nicksellen: Accidental comment "refactor"

tiltec:
Ah, need to move the migration.

foodsaving/groups/tests/test_data_migrations.py
```diff
@@ -2,10 +2,30 @@
...
+    def setUpBeforeMigration(self, apps):
+        User = apps.get_model('users', 'User')
+        Group = apps.get_model('groups', 'Group')
+        GroupMembership = apps.get_model('groups', 'GroupMembership')
+
+        user = User.objects.create()
+        group = Group.objects.create()
+        GroupMembership.objects.create(user=user, group=group)
+
+    def test_extract_pickups_from_stores_app(self):
```
tiltec: Test name doesn't quite fit ;)

tiltec:
I went through the PR in more detail and there are some bits missing or incorrect:

- `modify_notification_types` doesn't check for approval
- `AgreementViewSet` doesn't check for approval
- `initialize_group` will add the oldest member as membership manager, even if they are not approved (anymore)
- `leave_group_handler` (to remove user from pickups) should also be triggered when the approval role gets removed
- `subscriptions/receivers.py` has lots of handler that send websocket notifications to `group.members.all()`, but it should only send to approved members

Additionally, it would be good to have timestamps when a group member got a role and when it was removed. The history table might serve this purpose.

@NerdyProjects are you motivated to continue working on this? Would be cool, but I'm not in a hurry to get it merged and more interested in getting the bigger picture clear -> https://github.com/yunity/karrot-frontend/issues/894

I also have in mind that @nicksellen had strong resistance against using the `membership` relation to store information about users who are not members yet (and possibly never become members). Is your resistance still standing?
Our reasoning for this architecture was to have only one "user <-> group" relationship table. But maybe there's a better option?

nicksellen:
> I also have in mind that @nicksellen had strong resistance against using the membership relation to store information about users who are not members yet (and possibly never become members). Is your resistance still standing?

Ah, I think I did not express what my resistance was too clearly enough. It was more about wording/concepts/terminology than database tables, well, the interaction between them. I'm not proposing any change to the database structure.

I don't like the idea that a person could have a membership entry in the database, but not be called a member. It would meant there are two concepts of what "membership" means: the database level (an entry in the membership table), and the application level (an entry in the database with an "approved" role, or maybe different things in the future). This seems unclear and has the potential for confusion.

It might be the frontend would have places that say "members" but only show "approved members", but also might have other buttons/text nearby that clarify it (as people might be looking for other types of members, like inactive, or ones seeking approval.

NerdyProjects:
I would love to have a bigger picture clearer so I can more reasonable do changes - but all your suggestions seem reasonable to me, @tiltec .

I am happy to continue fixing the mentioned issues, but I might take some days.

tiltec:
Maybe we could have a talk about it at the end of this week in person? Some topics that come to my mind:

- Will there be a special "requested approval" state for new members or is it implicit when there's no approval role in the relation table?
- Naming of `group.membership`
- How does it interact with the wider topic of https://github.com/yunity/karrot-frontend/issues/853 ? - could become problematic if only one member needs to approve...

tiltec:
Group approval will be implemented in a different way in #561 
The methods explored in this PR might become handy when implementing UL2 (aka _student mode_) as outlined in https://github.com/yunity/karrot-frontend/issues/1062

PR #521, opened 2018-03-04
Add unsubscribe link to summary emails

nicksellen:
... also increased font-size of message content


foodsaving/base/templates/_base-head.mjml
```diff
@@ -36,10 +36,6 @@
         color: #4a982e;
     }
 
-    .message-content {
```
tiltec: This possibly changes the formatting of the message notification e-mail in an unwanted way (especially with markdown headlines). Maybe we shouldn't use the same class.

PR #524, opened 2018-03-05
Use yapf for consistent formatting

nicksellen:
You can run it like this:

```
yapf -i -r -e 'foodsaving/*/migrations'  foodsaving
```

... and there is an intellij plugin.

### DIFF
--- a/foodsaving/conversations/api.py
+++ b/foodsaving/conversations/api.py
@@ -10,18 +10,11 @@ from rest_framework.permissions import IsAuthenticated, BasePermission
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
-from foodsaving.conversations.models import (
-    Conversation,
-    ConversationMessage,
-    ConversationMessageReaction
-)
+from foodsaving.conversations.models import (Conversation, ConversationMessage, ConversationMessageReaction)
 from foodsaving.conversations.serializers import (
-    ConversationSerializer,
-    ConversationMessageSerializer,
-    ConversationMessageReactionSerializer,
-    ConversationMarkSerializer,
-    ConversationEmailNotificationsSerializer,
-    EmojiField)
+    ConversationSerializer, ConversationMessageSerializer, ConversationMessageReactionSerializer,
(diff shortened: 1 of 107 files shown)


nicksellen:
Sometimes it makes multiline things go onto one line, but because I configured `split_arguments_when_comma_terminated` you can make it split to multiple lines by having a trailing comma.

nicksellen:
Obviously, this cannot be merged, this is just for seeing whether we like it.

tiltec:
Yay for autoformatting :)

foodsaving/conversations/api.py
```diff
@@ -82,20 +74,16 @@ def email_notifications(self, request, pk=None):
         return Response(serializer.data)
 
 
-class ConversationMessageViewSet(
-    mixins.CreateModelMixin,
-    mixins.ListModelMixin,
-    GenericViewSet
-):
+class ConversationMessageViewSet(mixins.CreateModelMixin, mixins.ListModelMixin, GenericViewSet):
```
tiltec: I think we should go for the multiline version, putting a comma after the last entry.

tiltec:
Maybe now is a good time to integrate yapf?

nicksellen:
Sounds good. Might also be able to just use the _chromium_ preset, I used it for something else, and it seemed to fit what we like, didn't check in detail though.

Might need a totally branch on-top of current master. Or maybe just formatting master, then rebasing could work. Who knows.

nicksellen:
Ok, I reworked this now on-top of master, and included `./yapf.py` script to run it (with our excludes).

In the future if https://github.com/google/yapf/pull/543 gets merged we can use a yapf ignore file instead.

So, the plan would be:
- merge this to master
- cherry pick the "add yapf config" commit into any PR branch you might want to merge in the future
- run `./yapf.py` on your branch
- hopefully when you come to merge it'll all work nicely :)

Until we have gone through all the active branches like this, we should not manually reformat too much, or change the yapf config... keeps the pain down a bit. After that maybe some cleanup phases to (for example add lots of trailing commas).

PR #525, opened 2018-03-07
Add pickup notification emails

nicksellen:
Ok, basic functionality is there, todo:

- [x] check it manually a bit more
- [x] add tests


foodsaving/groups/templates/group_summary.mjml
```diff
@@ -59,7 +59,7 @@
                     <mj-divider/>
                     <mj-text font-size="13px" line-height="18px" color="#777">
                         <p>
-                            {% trans %}You are receiving this because you are subscribed to <strong>Weekly summary</strong> notifications for this group.{% endtrans %}
+                            {% trans %}You are receiving this because you are subscribed to <strong>Weekly summary</strong> emails for this group.{% endtrans %}
```
tiltec: I like "e-mails" instead of "email" in written text

nicksellen: http://grammarist.com/style/e-mail-email/ says:

> A few editorially conservative publications still prefer e-mail to email, but most of the English-speaking world has adopted the unhyphenated form. In a Google News search covering 2011 and the start of 2012, there are approximately six instances of email for every e-mail, a dramatic shift from a couple of years ago. And the unhyphenated form is even more common outside newswriting.

I get the impression the hypenated version is used in German more?

tiltec: Ah yes, that might be the reason... well, onto more consistency in future.

foodsaving/groups/serializers.py
```diff
@@ -274,7 +274,10 @@ def update(self, instance, validated_data):
 
 class GroupMembershipAddNotificationTypeSerializer(serializers.Serializer):
     notification_type = serializers.ChoiceField(
-        choices=(GroupNotificationType.WEEKLY_SUMMARY,),
+        choices=(
```
tiltec: Ah, much easier than expected. I guess that also works for model fields: "An iterable (e.g., a list or tuple) consisting itself": https://docs.djangoproject.com/en/2.0/ref/models/fields/

PR #526, opened 2018-03-08
Add group status field (replaces 'active' field)

tiltec:


tiltec:
The goal was to use it as CHOICES field. Maybe we can use the class style and autoconvert to choices style?

nicksellen:
Yes, seems possible to iterate over the constants inside the class with:
```py
vars(SomeClass)
```

... but that includes lots of internal things starting with underscores, which would need to be filtered out, and alternative would be using Enum, with string values. They support proper interation.

```py
class Foo(Enum):
    YAY = 'yay'
    WOO = 'woo'
```

If so, would also switch over the GroupNotificationType to use that style.

PR #528, opened 2018-03-08
WIP: Automated member removal

nicksellen:
I removed the membership removal code from https://github.com/yunity/karrot-backend/pull/509 and put it in here, so if we revive it we have the logic, the templates, etc.


tiltec:
Having talked to Karrot users, it seems reasonable to leave people in the group even if they are inactive for quite a long time. They might enjoy other activities such as travelling and would probably be unhappy if they find themselves without a group in karrot when they return.

But what is "quite a long time"? I propose at least 3 months, maybe even up to 6 months. That already includes one month where they are still shown as active.

On the other hand, groups might find it a bit unsettling if many inactive people are in the group. To reduce confusion, we should make it clear what the process for automated removal is.

It would be nice to send the "group is missing you" email again one month before the user gets removed from the group. The behavior would be:

- one month inactive: "group is missing you" and mark as inactive
- 3 months inactive: "group is missing you" again
- 4 months inactive: removal from group

Regarding removal from gropu: it seems a bit pointless to inform the user that they have been removed from a group when it's too late already. Instead of sending an email for "removal from group", I think it would be better to just inform them when they visit Karrot again.

tiltec:
I talked with Nick about it, we agreed on waiting a few months to see what users think about the "mark as inactive" feature before going forward with member removal.

PR #529, opened 2018-03-08
Don't notify about non-active store pickups

nicksellen:
The first pickup notifications would notify about pickups in non-active stores, this changes it so it matches the frontend logic to hide pickups unless the store is `active`.

### DIFF
--- a/foodsaving/stores/models.py
+++ b/foodsaving/stores/models.py
@@ -1,15 +1,24 @@
+from enum import Enum
+
 from django.conf import settings
 from django.db import models
 
 from foodsaving.base.base_models import BaseModel, LocationModel
 
 
+class StoreStatus(Enum):
+    CREATED = 'created'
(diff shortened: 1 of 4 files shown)


foodsaving/stores/models.py
```diff
@@ -1,15 +1,24 @@
...
 
 from foodsaving.base.base_models import BaseModel, LocationModel
 
 
+class StoreStatus(Enum):
+    CREATED = 'created'
+    NEGOTIATING = 'negotiating'
+    ACTIVE = 'active'
+    DECLINED = 'declined'
+    ARCHIVED = 'archived'
```
nicksellen: I tried the Enum style, it seems ok, the most annoying thing is having to put `.value` on the end to get the actual string out, if you forget that when you pass it into a `.filter()` call, then nothing will match.

PR #530, opened 2018-03-09
Add feedback to weekly summaries

nicksellen:
I noticed there is some nice feedback in some of the groups I am in, so should show it in the summary email to for motivating factors.

TODO:
- [x] add stats collection to make sure the emails don't get too enormous (messages + feedback particularly)

Looks a bit like this:

summarywithfeedback


nicksellen:
Ok, should be good. Would be great to get it on at least dev.karrot.world by Sunday morning to see them sent out :)

PR #532, opened 2018-03-11
WIP: Upgrade to django channels 2

nicksellen:
It was actually quite easy to get the app working with channels2, the tricky bit came getting the tests to work.

It sorta-kinda works but I think there is a better way.... I just don't know exactly what it is for now. Maybe moving to pytest properly as https://channels.readthedocs.io/en/latest/topics/testing.html wants us to.

I added http://asynctest.readthedocs.io/en/latest/ to support async in the tests that need it (I also notice that https://github.com/alexhayes/django-async-test exists which builds on top of it).

I used it with the `use_default_loop = True` option, which makes it use the "main" loop (instead of creating one during each test) and I think this might be the cause of some of my troubles (as I don't think things get cleaned up properly, leading to cascading test failures sometimes).

Anyway, too sleepy now.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -202,13 +202,14 @@ FEEDBACK_POSSIBLE_DAYS = 30
 
 CHANNEL_LAYERS = {
     "default": {
-        "BACKEND": "asgi_redis.RedisChannelLayer",
+        "BACKEND": "channels_redis.core.RedisChannelLayer",
         "CONFIG": {
             "hosts": [(REDIS_HOST, 6379)],
         },
-        "ROUTING": "foodsaving.subscriptions.routing.channel_routing",
     },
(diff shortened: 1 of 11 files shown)


nicksellen:
I tried out pytest a little, and the in memory channel layer, but the version in 2.0.2 is not really complete yet, and my tests just hang (similar to https://github.com/django/channels/issues/882#issuecomment-363940977) - I think https://github.com/django/channels/pull/863 addresses it, but it's not released yet.

So, I think we should just wait a bit until a few more releases of it. I don't think we have any pressing reason to upgrade anyway.

nicksellen:
Re-implemented in https://github.com/yunity/karrot-backend/pull/565

PR #533, opened 2018-03-11
Handle timezone in jinja2 filters

nicksellen:
The babel datetime library is not connected with django so it does not know to get the current timezone from there.

Should fix it :)

### DIFF
--- a/foodsaving/utils/email_utils.py
+++ b/foodsaving/utils/email_utils.py
@@ -7,6 +7,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.template import TemplateDoesNotExist
 from django.template.loader import render_to_string, get_template
 from django.utils import translation
+from django.utils.timezone import get_current_timezone
 from django.utils.translation import to_locale, get_language
 from furl import furl
 from jinja2 import Environment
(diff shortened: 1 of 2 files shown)


foodsaving/utils/email_utils.py
```diff
@@ -18,11 +19,20 @@
 
 
 def date_filter(value):
-    return format_date(value, format='full', locale=to_locale(get_language()))
+    return format_date(
+        value.astimezone(get_current_timezone()),
```
tiltec: Seems weird that `format_date` doesn't support the `tzinfo` parameter. Is this equivalent?

foodsaving/utils/tests/test_email_utils.py
```diff
@@ -1,14 +1,18 @@
...
 from foodsaving.groups.factories import GroupFactory
 from foodsaving.invitations.models import Invitation
 from foodsaving.userauth.models import VerificationCode
 from foodsaving.users.factories import UserFactory
 from foodsaving.utils import email_utils
+from foodsaving.utils.email_utils import time_filter, date_filter
 
 
-class TestEmailUtils(APITestCase):
+class TestEmailUtils(TestCase):
```
tiltec: It seems the tests fail because `pytest` doesn't allow a `unittest.TestCase` to access the database, but a `django.test.TestCase`: https://pytest-django.readthedocs.io/en/latest/helpers.html#pytest-mark-django-db-transaction-false-request-database-access

nicksellen: Ah, makes sense!

foodsaving/utils/email_utils.py
nicksellen: Yes, does seem odd. I think the logic was "oh, it's a date, not a time, so it doesn't need a timezone".

But _astimezone_ handles this fine. I think the function just doesn't care about the timezone, and asks for hour/minute/etc.

So, this is still a datetime with a timezone, but with the actual timezone, not utc. The point in time that it represents is still the same, just a different timezone perspective on it.

PR #534, opened 2018-03-11
Don't send notifications to inactive users

nicksellen:
Inactive users should not be notified about 3 kinds of things:
- group wall messages
- daily pickup notifications
- weekly group summary emails

They should still receive other emails.

### DIFF
--- a/foodsaving/conversations/tests/test_api.py
+++ b/foodsaving/conversations/tests/test_api.py
@@ -167,8 +167,7 @@ class TestConversationsEmailNotificationsAPI(APITestCase):
     def setUp(self):
         self.user = VerifiedUserFactory()
         self.group = GroupFactory(members=[self.user])
-        self.conversation = Conversation.objects.get_or_create_for_target(self.group)
-        self.conversation.join(self.user)
+        self.conversation = self.group.conversation
         self.participant = ConversationParticipant.objects.get(conversation=self.conversation, user=self.user)
 
     def test_disable_email_notifications(self):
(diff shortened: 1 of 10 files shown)


foodsaving/conversations/tests/test_api.py
```diff
@@ -211,13 +210,14 @@ def test_enable_email_notifications(self):
 
     def test_send_email_notifications(self):
         users = [VerifiedUserFactory() for _ in range(3)]
-        [self.conversation.join(u) for u in users]
+        [self.group.add_member(u) for u in users]
```
nicksellen: Checking the active-ness of the group member also means a by-product is we check they are a member at all, so they can't just be in the conversation as was the case before in the test.

foodsaving/groups/tasks.py
```diff
@@ -68,8 +69,8 @@ def send_summary_emails():
                     email.send()
                     email_count += 1
                     email_recipient_count += len(email.to)
-                except Exception:
-                    traceback.print_exc()
+                except AnymailAPIError:
+                    sentry_client.captureException()
```
nicksellen: Unrelated change I snuck in, seems a better approach I borrowed from another place in the code.

foodsaving/conversations/tests/test_tasks.py
```diff
@@ -0,0 +1,25 @@
+from django.core import mail
+from django.test import TestCase
+from django.utils import timezone
+
+from foodsaving.conversations.tasks import notify_participants
+from foodsaving.groups.factories import GroupFactory
+from foodsaving.users.factories import VerifiedUserFactory, UserFactory
+
+
+class TestConversationNotificationTask(TestCase):
```
nicksellen: I wrote this test before noticing there was some related ones inside the api test file, which did not seem the right place for them. This is about running the task function itself, so maybe it's seperate...

PR #536, opened 2018-03-19
Rework weekly summary

tiltec:
- Don't send empty summary emails
- Add name to feedback
- Add colon before member list

In my opinion, it would make sense to show a timestamp next to messages and the pickup time next to feedback. What do you think about it?

### DIFF
--- a/foodsaving/groups/emails.py
+++ b/foodsaving/groups/emails.py
@@ -56,7 +56,7 @@ def prepare_group_summary_data(group, from_date, to_date):
         group_id=group.id,
     )
 
-    return {
+    data = {
         # minus one second so it's displayed as the full day
         'to_date': to_date - relativedelta(seconds=1),
         'from_date': from_date,
@@ -68,6 +68,11 @@ def prepare_group_summary_data(group, from_date, to_date):
         'messages': messages,
         'settings_url': settings_url,
     }
+
+    data['has_activity'] = any(data[field] > 0 for field in ['pickups_done_count', 'pickups_missed_count']) or \
+        any(len(data[field]) > 0 for field in ['feedbacks', 'messages', 'new_users'])
+
+    return data
 
 
 def prepare_group_summary_emails(group, context):
--- a/foodsaving/groups/tasks.py
+++ b/foodsaving/groups/tasks.py
@@ -69,15 +69,17 @@ def send_summary_emails():
             email_recipient_count = 0
 
             context = prepare_group_summary_data(group, from_date, to_date)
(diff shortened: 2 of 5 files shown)


foodsaving/groups/emails.py
```diff
@@ -69,6 +69,11 @@ def prepare_group_summary_data(group, from_date, to_date):
         'settings_url': settings_url,
     }
 
+    data['has_activity'] = any(data[field] > 0 for field in ['pickups_done_count', 'pickups_missed_count']) or \
+        any(len(data[field]) > 0 for field in ['feedbacks', 'messages', 'new_users'])
```
nicksellen: It would seem slightly nicer to filter for no activity higher up with the orm objects rather than the output data structures so there is more certainty that the types it is checking is correct.

foodsaving/groups/emails.py
```diff
@@ -69,6 +69,11 @@ def prepare_group_summary_data(group, from_date, to_date):
...
+
+    return data
```
nicksellen: Returning None when there is no useful data seems more useful?

tiltec: I think it would be nice to have the stats written to influxdb even if no emails is sent. This was the easiest option, otherwise the calling function would need to do more work.

nicksellen: I see I see. That is somewhat repurposing the intention of that stat (`karrot.email.group_summary`), it's for tracking emails that were sent (with their details), sending an email with 0 recipients is not really a thing. Is there anything in it that you cannot fetch from influxdb already?

tiltec: I think the only purpose is to check if the task actually ran, but there's already the other stat `group__send_summary_emails`. Ok, so it seems better to not send out the `karrot.email.group_summary` stat.

nicksellen: One of the original purposes was to keep track of the point at which we send too much information in that email. At the moment it shows _all_ the messages/feedback.

Another purpose is to have a way to assess if the logic is correct (e.g. no sudden leap in the amount of things matched), so we can be slightly better positioned to detect that.

It also might be nice to know how many people are being sent the weekly notifications.

Checking if the task ran is not the purpose of that stat.

tiltec: > no sudden leap in the amount of things matched

Wouldn't it be helpful to see when no things matched (instead of no data point)?

nicksellen: Ok, lets go for it.

PR #538, opened 2018-03-28
Ignore deleted pickups in emails

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/987


nicksellen:
Should be good to go now :)

PR #541, opened 2018-04-16
Edit message endpoints

jskollin:
One test case: _test_receives_messages_, continues to fail. We added the parsing recommended by @tiltec, however the test is still failing. We investigated the issue and are not sure of the origin.


tiltec:
Regarding the failing test: there's another place where the message serializer is checked. You need to add the `updated_at` field as well:

in `ConversationReceiverTests.test_receives_messages`
```py
        # author should get message & updated conversations object too
        response = author_client.receive(json=True)
        response['payload']['created_at'] = parse(response['payload']['created_at'])
        self.assertEqual(response, {
            'topic': 'conversations:message',
            'payload': {
                'id': message.id,
                'content': message.content,
                'author': message.author.id,
                'conversation': conversation.id,
                'created_at': message.created_at,
                'received_via': '',
               'reactions': []
            }
        })
```

I guess this should be abstracted at some point, seems a bit annoying to update it at two places...

jskollin:
I fixed the failing test case and added the check to limit the editing period.

.gitignore
```diff
@@ -40,3 +40,4 @@ uploads
 
 # Mac files
 .DS_Store
+foodaving/.DS_Store
```
tiltec: Maybe `**/.DS_Store` would be a better expression as it will match in all directories?
https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.2.txt#L85

Also, you need to `git rm` the DS_Store files that you already added.

foodsaving/conversations/api.py
```diff
@@ -55,6 +58,27 @@ def has_object_permission(self, request, view, message):
...
+class IsAuthorConversationMessage(BasePermission):
+    """Is the user the author of the message they wish to update?"""
+
+    message = _('You are not the author of this message')
+
+    def has_object_permission(self, request, view, message):
+        return request.user == message.author
+
+
+class IsWithinUpdatePeriod(BasePermission):
```
tiltec: Could you add one test where editing fails if the period is over? Similar to https://github.com/yunity/karrot-backend/blob/951c911eef639ee1ef43f7e7f8ae22ef106021c1/foodsaving/pickups/tests/test_feedback_api.py#L310

jskollin:
I added the test case: test_cannot_update_message_if_past_10_days! Thanks :)

PR #542, opened 2018-05-20
fixes #1011

pogopaule:
Closes https://github.com/yunity/karrot-frontend/issues/1011

### DIFF
--- a/foodsaving/users/models.py
+++ b/foodsaving/users/models.py
@@ -143,6 +143,9 @@ class User(AbstractBaseUser, BaseModel, LocationModel):
     def _send_mail_change_notification(self):
         prepare_changemail_success_email(self).send()
 
+    def resend_mail_verification_code(self):
+        self._send_welcome_mail()
+
     @transaction.atomic
     def _send_welcome_mail(self):
         self._unverify_mail()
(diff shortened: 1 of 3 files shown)


foodsaving/users/models.py
```diff
@@ -143,6 +143,9 @@ def update_language(self, language):
     def _send_mail_change_notification(self):
         prepare_changemail_success_email(self).send()
 
+    def resend_mail_verification_code(self):
+        self._send_welcome_mail()
```
tiltec: This causes unexpected behavior if the user already has a verified email address and then requested to change their email address. If the user would now request to resend the verification code, they would get an "welcome" email and the email gets marked as "not verified".

We could add a check if `user.mail_verified` is true: if yes, then run `send_mail_verification_code`, if not then `_send_welcome_mail` is appropriate.

Ping @mr-kojo to check this proposal as it diverts from the solution proposed in https://github.com/yunity/karrot-frontend/issues/1011

mr-kojo: Unfortunately, checking `user.mail_verified` doesn't work because it is `false` in both use cases.

However, I think we can distinguish the use cases by checking if `user.email == user.unverified_mail`. If `true`, then call `send_mail_verification_code`, otherwise call `_send_welcome_mail`.
@tiltec, do you agree?

tiltec: Seems good! As far as I remember, that's also the method I chose in the old frontend implementation.

PR #544, opened 2018-06-10
Add pickup conversations

nicksellen:
https://github.com/yunity/karrot-frontend/issues/677

Added pickup conversations to the backend. Basically just copied the implementation for the group conversations.

So, one new endpoint: `GET /api/pickup-dates/{id}/conversation/` to get the conversation for a pickup. Then you would use the message endpoints as used by the group wall for the rest.

I think we discussed before that we found this API style a bit odd in the frontend, as we kind of wanted to just have the conversation id included in the parent resource I think (i.e. a field in the group or pickup). So, perhaps I could rejig them both to try that style if you think it worth it.

### DIFF
--- a/foodsaving/conversations/api.py
+++ b/foodsaving/conversations/api.py
@@ -167,7 +167,7 @@ class RetrieveConversationMixin(object):
 
     def retrieve_conversation(self, request, *args, **kwargs):
         target = self.get_object()
-        conversation = Conversation.objects.get_for_target(target)
+        conversation = Conversation.objects.get_or_create_for_target(target)
         serializer = ConversationSerializer(conversation, data={}, context={'request': request})
         serializer.is_valid(raise_exception=True)
         return Response(serializer.data)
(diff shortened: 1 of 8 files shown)


foodsaving/conversations/api.py
```diff
@@ -168,5 +168,6 @@ class RetrieveConversationMixin(object):
     def retrieve_conversation(self, request, *args, **kwargs):
         target = self.get_object()
         conversation = Conversation.objects.get_or_create_for_target(target)
-        serializer = ConversationSerializer(conversation, context={'request': request})
+        serializer = ConversationSerializer(conversation, data={}, context={'request': request})
```
tiltec: This permission check seems to be duplicated. If the user isn't allowed to access the parent endpoint (e.g. group detail, pickup), then the queryset gets filtered and they also can't access the `/pickup/<id>/conversation` endpoint.

But maybe that's because the membership management is also duplicated in the conversation?

nicksellen: The conversation is only for people who are signed up to the pickup, so it's legitimate to have access to the pickup, but not the conversation.

nicksellen:
I found the reactive websocket updates were not setup for updating the other participants when someone joins/leaves a conversation, so have added that now.

tiltec:
Seems all necessary and good, merge as you like :)

PR #545, opened 2018-06-12
Fix sending group summary email fails if locale does not match yunity/karrot-frontend#1033

oldPadavan:


foodsaving/utils/email_utils.py
```diff
@@ -110,6 +111,11 @@ def prepare_email(template, user=None, context=None, to=None, language=None, **k
 
 
 def prepare_email_content(template, context, language='en'):
+    try:
+        Locale.parse(language)
```
tiltec: I think that doesn't work in all cases, because babel's `Locale.parse` accepts way more variants than Django. For example:

```py
> from babel.core import Locale
> print(Locale.parse('pt_BR'))
pt_BR

> from django.utils import translation
> print(translation.get_language_info('pt_BR'))
KeyError: 'Unknown language code pt_BR.'

> print(translation.get_language_info('pt-BR'))
{'bidi': False, 'code': 'pt', 'name': 'Portuguese', 'name_local': 'Português', 'name_translated': 'Portuguese'}
```

How about `translation.check_for_language(locale_code)` instead? It returns `True` if and only if the locale was found.

oldPadavan:
Thanks for pointing to the problem, I've modified the code for yunity/karrot-frontend#1033 according to Your advice

PR #546, opened 2018-06-13
Auto mark groups inactive #972

oldPadavan:
Implements yunity/karrot-frontend#972
I haven't implemented yet reactivation of groups. I see 2 options: add activation to the endpoint handlers for every event or to add it to overridden save method of Conversation, History and Feedback models. How would You suggest is better?


oldPadavan:
Thanks @tiltec, I think Your option is better one. My initial idea was to keep all checks of recent activity in one place - `Group` model, however anyway in order to reactivate the group it would be necessary to add some code to each of the events handlers, so it would be simpler to just update `last_active_at` as well in these handlers. I'll change the implementation

PR #548, opened 2018-06-19
Automatically mark groups inactive and reactivate them on actions

oldPadavan:
Implements yunity/karrot-frontend#972. Changed implementation according to comments from yunity/karrot-backend#546


foodsaving/conversations/api.py
```diff
@@ -190,6 +192,12 @@ def remove_reaction(self, request, pk, name):
         reaction.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    def perform_create(self, serializer):
+        message = serializer.save()
+        if message.conversation.target_type == ContentType.objects.get_for_model(Group):
```
tiltec: I lately used `isinstance(message.conversation.target, Group)`, but this seems fine as well. We could decide for one version to unify our styles :)

oldPadavan: Agree! Comparison using isinstance looks better. I've just committed updated version

PR #549, opened 2018-06-20
Disable creation of pickups if store is inactive

oldPadavan:
Don't create/modify pickups if store is not active - yunity/karrot-frontend#1026


tiltec:
This is good but doesn't address the problem of #1026 because the "create pickup" functionality is not exposed for non-active stores in the frontend.

I rather meant the automatic creation of pickup dates. This function: https://github.com/yunity/karrot-backend/blob/2aaf4c7610a088e8df54eed4d93d2b15d0cda25e/foodsaving/pickups/models.py#L24
It should only run for series that belong to active stores. Seems a pretty simple change actually!

oldPadavan:
Hm... it seems I didn't understand the problem. So we've got to just filter out inactive stores in `create_all_pickup_dates`?

PR #550, opened 2018-06-20
Prevent create_all_pickup_dates from generating for inactive stores

oldPadavan:
Instead of yunity/karrot-backend#549 for yunity/karrot-frontend#1026


tiltec:
I think now we need to run `series.update_pickup_dates()` when the store gets set back to `active` (for all series of the store).

PR #551, opened 2018-06-21
Enable pickup chat notification emails

tiltec:


foodsaving/conversations/emails.py
```diff
@@ -22,17 +35,73 @@ def prepare_conversation_message_notification(user, message):
...
-            from_email=from_email,
-            user=user,
-            reply_to=[reply_to],
-            context={
-                'conversation_name': conversation_name,
-                'author': message.author,
-                'message_content': message.content_rendered(),
-                'conversation_url': group_wall_url(group),
-                'mute_url': conversation_mute_url(group, message.conversation)
+    return prepare_email(
```
nicksellen: Why did you remove the translation override?

tiltec: It seems unnecessary, as `prepare_email` has its own translation override.

nicksellen: Ah, yes I see. Makes sense. And for the other one it does translation stuff in the preparation so it does need it.

PR #552, opened 2018-06-24
WIP: trusts implementation

alangecker:
first step of https://github.com/yunity/karrot-frontend/issues/878

implements the ability to give and revoke trust carrots which expire after a certain time.
the amount of trust carrots are limited per time to prevent abuse


foodsaving/users/models.py
```diff
@@ -217,3 +219,25 @@ def has_perm(self, perm, obj=None):
...
+
+
+class TrustManager(Manager):
+    def valid(self):
+        return self.filter(valid_from__gte=timezone.now() - relativedelta(days=settings.TRUST_VALID_TIME))
+
+
+class Trust(BaseModel):
+    objects = TrustManager()
+    user = ForeignKey(User, on_delete=models.CASCADE, related_name='trusts')
```
tiltec: This could use a UniqueTogether constraint on `(user, given_by`) to keep the database in good shape. Have a look at the Feedback model and serializer.

PR #554, opened 2018-06-27
Performance improvements

nicksellen:
Two things!

1. adding etag caching - saves a bit of bandwidth, probably not a big improvement, but why not!
1. not using `post_init` triggers - this is way too excessive, and was really just a crappy way of doing a data migration

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -249,6 +249,10 @@ HUEY = {
     'always_eager': True,
 }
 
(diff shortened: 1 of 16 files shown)


nicksellen:
Also added silk profiler behind USE_SILK environment variable, so `USE_SILK=true ./manage.py runserver` (also will have to run migrations first).

PR #563, opened 2018-07-17
Conversation threads

nicksellen:
To do

- [x] fix link in notification, should point to `group/:id/messages/:id/replies` or similar
- [x] fix reply-via-email to thread
- [x] prevent moving message between threads
- [x] show in notification email that it was a reply?
- [ ] follow the conversation thread by adding a reaction?


foodsaving/conversations/serializers.py
```diff
@@ -100,14 +106,112 @@ class Meta:
...
+            'id',
+            'author',
+            'content',
+            'conversation',
+            'created_at',
+            'updated_at',
+            'reactions',
+            'received_via',
+            'is_editable',
+            'thread',  # ideally would only be writable on create
```
tiltec: indeed it is possible to move messages between threads. bug or feature? ;)

PR #567, opened 2018-07-28
Add batched conversation notification emails

tiltec:
This reduces the number of emails if messages are being sent in quick succession.

We wait for 5 minutes after the message has been created. If there was another message in the meantime, skip the task. If not, send out one email per participant that contains all not-yet-notified messages.

If a message was marked as seen in the meantime, don't send out notifications (hence one email per participant and the number of messages per participant may vary).

Maybe unwanted side effect: if there's lots of discussion activity, notifications might only be sent out after the activity died down again. But maybe that's also fine...

Todo: 
- [x] try out the schedule feature of huey
- [x] test more thread stuff (see codecov)
- [x] exclude replies from conversations


foodsaving/conversations/emails.py
```diff
@@ -17,76 +17,90 @@
...
-        if message.is_thread_reply():
-            return prepare_group_thread_message_notification(user, message)
-        return prepare_group_conversation_message_notification(user, message)
-    if isinstance(message.conversation.target, PickupDate):
-        return prepare_pickup_conversation_message_notification(user, message)
-    if isinstance(message.conversation.target, GroupApplication):
-        return prepare_group_application_message_notification(user, message)
-    if not message.conversation.target and message.conversation.is_private:
-        return prepare_private_user_conversation_message_notification(user, message)
+def nice_from_email_text(messages):
```
nicksellen: Grammar of the name confuses me, like the input is _email text_ and the output is some _nice_ ....

foodsaving/conversations/emails.py
```diff
@@ -17,76 +17,90 @@
...
+    return ', '.join(set(message.author.display_name for message in messages))
+
+
+def prepare_conversation_message_notification(user, messages):
+    assert len(messages) > 0
```
nicksellen: `assert` will not be active unless `__debug__` is True - https://docs.python.org/3/reference/simple_stmts.html#assert. I'm not sure what value we would have in production, but could be False.

foodsaving/template_previews/views.py
```diff
@@ -44,8 +44,12 @@ def shuffle_groups():
     return Group.objects.order_by('?')
 
 
-def random_message():
-    return ConversationMessage.objects.exclude_replies().order_by('?').first()
+def random_messages():
+    conversation = ConversationMessage.objects.order_by('?').first().conversation
+    messages = conversation.messages.exclude_replies().all()
+    if len(messages) > 5:
+        messages = messages[:4]
```
nicksellen: Why not a limit on the query?

foodsaving/conversations/tests/test_tasks.py
```diff
@@ -1,24 +1,122 @@
...
         self.user = VerifiedUserFactory()
-        self.unverified_user = UserFactory()
-        self.inactive_user = VerifiedUserFactory()
         self.author = VerifiedUserFactory()
-        self.group = GroupFactory(members=[self.author, self.user, self.inactive_user, self.unverified_user])
-        self.group.groupmembership_set.filter(user=self.inactive_user).update(inactive_at=timezone.now())
-        self.message = self.group.conversation.messages.create(author=self.author, content='foo')
+        self.group = GroupFactory(members=[self.author, self.user])
         mail.outbox = []
+        with patch('foodsaving.conversations.receivers.tasks.notify_participants'):
```
nicksellen: This is used quite a few times, and could be clearer to refactor into:
```py
with suppressed_notifications():
  # foo
```

foodsaving/conversations/tests/test_tasks.py
```diff
@@ -1,24 +1,122 @@
...
+        self.assertEqual(len(mail.outbox), 0)
+
+    def test_exclude_already_notified_in_thread(self):
+        self.group.conversation.messages.create(author=self.user, thread=self.message, content='first thread reply')
+        mail.outbox = []
+        self.group.conversation.messages.create(author=self.user, thread=self.message, content='second thread reply')
+
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn('second thread reply', mail.outbox[0].body)
+        self.assertNotIn('first thread reply', mail.outbox[0].body)
```
nicksellen: Maybe a test with multiple participants (in different states of notified/read) would be useful to make sure they each get the correct notification.

foodsaving/conversations/tasks.py
```diff
@@ -34,10 +46,15 @@ def notify_participants(message):
             user__groupmembership__group=message.conversation.target,
         )
 
-    for participant in participants_to_notify:
+    for participant in participants_to_notify.distinct():
+        messages = participant.unseen_and_unnotified_messages().all()
```
nicksellen: Looks like we didn't filter for participants on their notified_up_to, so could return 0 messages here? (would be the case if this is run in quick succession I think)

foodsaving/conversations/tasks.py
```diff
@@ -22,8 +30,12 @@ def notify_participants(message):
             email_notifications=True,
         )
 
-    participants_to_notify = participants_to_notify.exclude(user=message.author).exclude(
+    participants_to_notify = participants_to_notify.exclude(
+        user=message.author,
+    ).exclude(
         user__in=User.objects.unverified_or_ignored(),
+    ).exclude(
+        seen_up_to=message,
```
nicksellen: Greater or equal than this message would be more accurate?

foodsaving/template_previews/views.py
tiltec: It is a limit on the queryset: https://docs.djangoproject.com/en/dev/topics/db/queries/#limiting-querysets
But I need to check for the length anyways, as there might be less messages than requested. Maybe the limit is not needed at all though. But then the email preview might get a bit long

foodsaving/conversations/tasks.py
tiltec: I added a filter for notified_up_to. I guess it still could happen that a messages is being marked as seen while the task is running. This might result in 0 messages, so I could add another filter... or just wait if this happens in production. It will throw an error, get reported by sentry and not affect anybody..

foodsaving/template_previews/views.py
nicksellen: But `len(messages)` will evaluate the query then count the entries, no?

PR #572, opened 2018-08-05
Improve message extraction

tiltec:
Djangos localization message extraction is quite lacking and seems quite hard to extend (django-jinja has to [monkey-patch some regular expressions](https://github.com/niwinz/django-jinja/blob/master/django_jinja/management/commands/makemessages.py)).
I found out today that our current setup does not support the new [`trimmed` jinja2 parameter](http://jinja.pocoo.org/docs/2.10/templates/#i18n). I then switched to [extracting messages with babel](http://babel.pocoo.org/en/latest/cmdline.html), which seems the way to go.

I took the chance to customize and add some new management commands:

- makemessages
- pushmessages
- pullmessages
- compilemessages


foodsaving/locale/en/LC_MESSAGES/django.po
```diff
@@ -440,6 +439,13 @@ msgstr ""
 msgid "%(group_name)s upcoming pickups"
 msgstr ""
 
+#: foodsaving/pickups/templates/pickup_notification.html.jinja2:130
+#, python-format
+msgid ""
+"                                                    "
+"%(empty_collectors_count)s/%(max_collectors)s remaining\n"
```
tiltec: This was actually completely missed by the previous extraction method...

PR #573, opened 2018-08-19
Conversation overview API

tiltec:
Related to https://github.com/yunity/karrot-frontend/issues/1070
Frontend PR: https://github.com/yunity/karrot-frontend/pull/1084

What is this all about? -> https://github.com/yunity/karrot-frontend/issues/1070#issuecomment-414126846

# What does this PR do?

- changes response type of `GET /api/conversations/` to include related messages, pickups and applications and adds pagination
- adds new endpoint `GET /api/messages/my_threads/` to list recently active threads the user is participant in
- adds `/api/conversations/?exclude_wall=True` filter to ignore group wall conversation
- adds `/api/conversations/?group=<id>` to filter by group, but always include private messages
- adds `/api/messages/my_threads/?group=<id>` to filter by group (looks up group conversation id)
- adds `Conversation.group` column for easier filtering
- adds `Conversation.latest_message` and `ConversationMessage.latest_message` columns for easier retrieval of latest conversation and thread messages
- adds lots of prefetching to make `GET /api/conversations/` and `GET /api/messages/my_threads/` reasonably efficient
- adds `Conversation.type()` method to abstract a bit over the content type and return it to the client
- fixes caching problems with `ConversationMessage._replies_count` and `_participant` in ConversationSerializer


tiltec:
Given I want to filter conversations by group, that's not possible with generic relations. My proposal would be to add a column `group` to the `Conversation` model that allows easy filtering. The column can also be `null` for private messages.

tiltec:
For delivering related data, such as pickups or applications (for pickup chats and application chats) or even just the latest message, it could be worth introducing a response structure like this:

```json
{
  conversations: [
    { id: 1, latest_message: 234, type: 'pickup', target_id: 5 },
    { id: 2, latest_message: 200, type: 'application', target_id: 99 },
  ],
  messages: {
    234: { id: 234, content: 'hello', ... },
    200: { id: 200, content: 'bla', ... },
  },
  pickups: {
    5: { id: 5, date... },
  },
  applications: {
    99: { id: 99, answers... },
   }
}
```

Slack API has a similar approach and it seems useful to me.

Currently, the conversation overview and the feedback list would benefit from this kind of API.  Maybe in future also users could be delivered like this (instead of loading all upfront).

@nicksellen what are your thoughts on this?

tiltec:
Ok, next iteration of conversations overview API pushed! More text tomorrow :) I should also adjust the frontend implemention to get a feeling if the new API structure fits well.

PR #585, opened 2018-11-07
WIP: Rename "store" to "place"

nicksellen:
As part of https://github.com/yunity/karrot-frontend/issues/1127

This is just the first part of exploring an approach.

I looked a little into just renaming the model:
- https://docs.djangoproject.com/en/2.1/ref/migration-operations/#renamemodel
- https://stackoverflow.com/q/25091130

But, it's also about renaming the whole app which seems very tricky staying within Django (https://gist.github.com/jamesmfriedman/6168003). Therefore I think an approach similar to https://stackoverflow.com/a/8408131.

The idea is just rewrite everything as if they were never called stores. Would be a heavy one-off deployment/migration to run, but my feeling is it might be simpler/cleaner than battling Django MIgrations (and probably having to leave some files in `foodsaving/stores`).

For this I used [codemod](https://github.com/facebook/codemod) to replace various bits in the file. In the end I didn't actually use it's manual check features so sed would probably have been fine. You have to install it independently (`pip2 install codemod` for me) and the run `./codemod.sh`. The second commit is the result of doing it.


nicksellen:
Things that all seem to work fine:
- migrations
- tests
- sample data
- the actual server

How to actually migrate an existing database with data is something I haven't looked into much right now....

nicksellen:
... also gunna try a bit more with Django

nicksellen:
Ok, I have a db script too now! So steps to try it out:
* I'd recommend to copy a db with data already in it so you can try out the migration easily (see below), afterwards switch local settings to use that db
* run `./db-rename-stores-to-places.py` and inspect the sql
* run `./db-rename-stores-to-places.py --execute` to run it against the db
* make sure everything works!
  * `./manage.py runserver` should not show any migation warnings
  * `./manage.py migrate` should be thoroughly tricked and believe everything is fine
  * `./manage.py test --parallel 4` should also be happy of course


```
# to copy fstool-db to fstool-copy
createdb -U postgres fstool-copy
pg_dump -U postgres fstool-db | psql -U postgres fstool-copy
```

foodsaving/locale/ar/LC_MESSAGES/django.po
```diff
@@ -298,7 +298,7 @@ msgstr ""
 
 #: foodsaving/groups/templates/group_summary.html.jinja2:171
 #, python-format
-msgid "%(weight)s kg from <a href=\"%(store_url)s\">%(store_name)s</a>"
+msgid "%(weight)s kg from <a href=\"%(place_url)s\">%(place_name)s</a>"
```
tiltec: Changes of msgid will probably invalidate translations of affected messages. We should tell translators before we release.

foodsaving/groups/stats.py
```diff
@@ -95,17 +95,17 @@ def get_group_members_stats(group):
...
 
-    for entry in group.store.values('status').annotate(count=Count('status')):
-        # record one value per store status too
+    for entry in group.place.values('status').annotate(count=Count('status')):
+        # record one value per place status too
         fields['count_status_{}'.format(entry['status'])] = entry['count']
 
     return [{
-        'measurement': 'karrot.group.stores',
+        'measurement': 'karrot.group.places',
```
tiltec: We should run something in influxdb to migrate measurements, otherwise stats become even more unreliable ;)

nicksellen: Ah yes, influxdb db too. Also the tag names.

nicksellen: It would need exporting, modifying, and reimporting, the export is like this:

```
influx_inspect export -database karrot-dev -datadir /var/lib/influxdb/data -waldir /var/lib/influxdb/wal -out /tmp/karrot-dev-export
```

A bit of find/replace, then a reimport (see https://docs.influxdata.com/influxdb/v1.7/tools/shell/#import-data-from-a-file-with-import). Didn't try that one.

Doesn't seem too hard, and would prefer that to having incorrect "legacy" names in the stats.

nicksellen: I tried it out, exporting fstool on yuca and importing locally, seemed ok and didn't take very long either.

```
influx_inspect export -database fstool -datadir /var/lib/influxdb/data -waldir /var/lib/influxdb/wal -out /tmp/fstool_export -compress
```
... copy file to my laptop ...

```
zcat fstool_export | sed 's/store/place/g' > fstool_export_place
influx -import -path=fstool_export_place
```

Import only took about a minute. Could also rename the db to "karrot-world" whilst we're there...

Incidentally notice the warning:
```
[I] 2018-11-07T18:58:45Z WARN: 100% of max-values-per-tag limit exceeded: (100000/100000), db=fstool shard=164 measurement=django_request tag=full_path service=store
```

tiltec: See here for more details about the warning: https://github.com/bitlabstudio/django-influxdb-metrics/issues/27
The problem won't occur anymore in my fork of django-influxdb-metrics that we use since a few weeks.

nicksellen:
Closing in favour of https://github.com/yunity/karrot-backend/pull/654

PR #591, opened 2018-12-11
Improve handling of pickup series modifications v2

tiltec:
Context: https://community.foodsaving.world/t/better-change-handling-for-recurring-pickups/174

#587 turned out to have some problems, here's the new approach: we always keep non-empty pickups, but they can be _disabled_ individually.

1. changing a recurring pickup *keeps non-empty pickups*
a. if time, weekdays or any other part of the recurrence rule changes, create new pickups where necessary and delete empty non-matching pickups
b. if the description of the recurring pickup changes, rewrite descriptions of its existing pickups
c. if max slots of the recurring pickup changes, rewrite max slots of its existing pickups
1. changing the "weeks in advance" setting of a store keeps non-empty pickups

- Disabled pickups can be enabled again.
- Pickups that are part of a series cannot be moved.
- Series API returns dates that belong to the series (client can identify leftover pickups)

To do
- [x] move series updates from store serializer into receiver
- [x] fix commented out test in test_pickupdate_series_api
- [x] remove `last_changed_by` from API and other places (if not necessary)

Open questions:
- how to migrate existing changed pickups?
- email/push notifications about disabled pickups?
- do anything on group timezone change?


tiltec:
Talked with @djahnie yesterday evening: seems to complex to properly implement moved recurring pickups, users can adapt. It's more important that the users can understand the behavior of the software.

I added a data migration that makes upcoming moved pickups into one-time pickups and it creates a dummy pickup as part of the series at the original time to prevent new pickups from being created automatically.

PR #593, opened 2019-01-03
Only mention the `--more` flag if not passed in.

lwm:
Tiny one for the sample data loader.

If you passed `--more` and you still see the message recommending to
pass the option, you might start to doubt that it is doing what it is
supposed to be doing.


tiltec:
Thanks! Could be tricky to add a test here, because it would slow down the test suite quite a bit. Would be fine for me to go without test.

lwm:
Oh yes, you're right. Merge away then :bomb:

PR #594, opened 2019-01-03
Messages marked as read when joining a conversation.

lwm:
~~Still working on this but here's the current progress. Probably the `ConversationParticipant` logic which sets the `seen_up_to = last_message` could probably live in some function on one of the related models. Will take another look at it soon. Would be good to know if this is going in the right direction :)~~

Ready for a review.

Closes https://github.com/yunity/karrot-frontend/issues/1106.

### DIFF
--- a/foodsaving/applications/tests/test_receivers.py
+++ b/foodsaving/applications/tests/test_receivers.py
@@ -1,0 +1,22 @@
+from foodsaving.conversations.models import ConversationParticipant
+from foodsaving.groups.factories import GroupFactory
+from foodsaving.users.factories import UserFactory
+from rest_framework.test import APITestCase
+
+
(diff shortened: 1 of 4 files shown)


tiltec:
Did you consider creating one receiver for `ConversationParticipant` which sets `seen_up_to` on creation?

```py
@receiver(post_save, sender=ConversationParticipant)
def mark_as_read_on_join(sender, instance, created, **kwargs)
  if not created:
    return
  # now mark as read
  # ...
```

Could also be put in the model `save` method.

lwm:
> Did you consider creating one receiver for ConversationParticipant which sets seen_up_to on creation?

Nope but that's a great idea, I'll do that then :)

lwm:
OK, this is ready for a review! I've manually tested it on some wall conversations.

PR #599, opened 2019-01-05
Sort PickupCollector by sign up order

lwm:
Was still kinda confusing but hopefully got the order correct!

Closes https://github.com/yunity/karrot-frontend/issues/1157.

Refs: https://community.foodsaving.world/t/feature-request-keep-order-of-signing-up-to-pickups/185


tiltec:
I just had this test failing locally, but couldn't reproduce it. But it made me curious if the ordering would actually change, so I modified '-created_at' to 'created_at'. It didn't seem to make any difference to the API test, always passing (except for the one spurious case).

So it seems to me that setting the 'ordering' doesn't affect ManyToMany lookups and hence doesn't solve https://github.com/yunity/karrot-frontend/issues/1157 - the reason why the mostly passes seems to be the unspecified default ordering of the database. Which might also be the reason why users came to expect that the order was kept, where in fact we never made sure to really do so.

I will reopen the issue, to have a closer look at it.

lwm:
Oh damn, a flaky test! Sorry. So, here is what I did to make sure this is happening:

```
$ dcp exec backend env/bin/pip3 install pytest-repeat
```

And then, since `pytest-repeat` tells me "Repeating unittest class tests not supported", I made this diff:

```diff
+import pytest
+@pytest.mark.django_db
+def test_pickup_collectors_reverse_created_at_order():
+    first_member = UserFactory()
+    second_member = UserFactory()
+    group = GroupFactory(members=[first_member, second_member])
+    store = StoreFactory(group=group)
+    pickup = PickupDateFactory(store=store)
+
+    first_pickup_join_time = timezone.now()
+    with freeze_time(first_pickup_join_time):
+        first_pickup_join = PickupDateCollector.objects.create(user=first_member, pickupdate=pickup)
+    assert first_pickup_join.created_at == first_pickup_join_time
+
+    second_pickup_join_time = first_pickup_join_time + timedelta(hours=1)
+    with freeze_time(second_pickup_join_time):
+        second_pickup_join = PickupDateCollector.objects.create(user=second_member, pickupdate=pickup)
+    assert second_pickup_join.created_at == second_pickup_join_time
+
+    assert PickupDateCollector.objects.first() == second_pickup_join
+
+
+import pytest
+@pytest.mark.django_db
+def test_join_pickup_order_by_sign_up(client):
+    url = '/api/pickup-dates/'
+    member = UserFactory()
+    second_member = UserFactory()
+    group = GroupFactory(members=[member, second_member])
+    store = StoreFactory(group=group)
+    pickup = PickupDateFactory(store=store)
+    pickup_url = url + str(pickup.id) + '/'
+    join_url = pickup_url + 'add/'
+
+    client.force_login(user=second_member)
+    response = client.post(join_url)
+
+    client.force_login(user=member)
+    response = client.post(join_url)
+
+    response = client.get(pickup_url)
+    assert response.data['collector_ids'][0] == second_member.id
+    assert response.data['collector_ids'][1] == member.id
```

Then, running:

```
$ dcp exec backend pytest --count=100 -vvv -k test_pickup_collectors_reverse_created_at_order # no fails
$ dcp exec backend pytest --count=100 -vvv -k test_join_pickup_order_by_sign_up # fails a lot 
```

I get a number of failures on the API level test with the wrong ordering. AFAIK, we are fine on the model layer, since the ordering is honoured but on the relation, as you said, the ordering is disabled (found this loosely documented in a few places).

So, it seems the solution is:

  * Honour the ordering on the serializer (as described [in this SO answer](https://stackoverflow.com/questions/25134004/how-to-order-a-django-rest-framework-manytomany-related-field)).

WDYT?

tiltec:
I played a bit with the serializer method field and it seems the way to go. But there are "n+1 problems" with this, because they do a DB request per object to be serialized.
It seems to be possible to order prefetched relations: https://stackoverflow.com/questions/48071397/django-prefetch-relateds-prefetch-order-by

lwm:
Oh, that is good that we can order prefecthing. I can't seem to find any usage in the codebase, so I recommend using this `pytest-django` fixture to make sure you are getting the number of queries you expect: https://pytest-django.readthedocs.io/en/latest/helpers.html#django_assert_num_queries. I can also do this, since I made the mess :)

tiltec:
I sometimes used [assertNumQueries](https://docs.djangoproject.com/en/2.1/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) from Djangos TransactionTestCase, which seems to be equivalent. For some reason, I keep using _unittest_ style, although we have some pytest tests for the async bits.

tiltec:
I did a bit more trial & error and the problem can be solved by using `pickupdatecollector_set` instead of the `collectors` M2M relation. Will push a branch!

PR #600, opened 2019-01-05
New members get editor role for playground groups.

lwm:
Not sure if we want to add more roles?

Closes https://github.com/yunity/karrot-frontend/issues/1140.


foodsaving/groups/tests/test_receivers.py
```diff
@@ -79,3 +79,15 @@ def test_non_send_group_join_stats_on_update(self, write_mock):
         membership.inactive_at = None
         membership.save()
         self.assertFalse(write_mock.called)
+
+
+class TestGroupPlaygroundReceivers(TestCase):
+    def setUp(self):
+        self.member = UserFactory()
+        self.first_member = UserFactory()
```
nicksellen: Seems unused?

nicksellen:
There's not so many roles used, so this should be good!

foodsaving/groups/tests/test_receivers.py
lwm: Yep, will fixup!

PR #604, opened 2019-01-06
Sort pickup collectors by date

tiltec:
`collectors` doesn't honor the ordering, but `pickupdatecollector_set` solves it. 

I added a bit more to the API test that reverses the order, to test both ways and be more sure we don't rely on unspecified database order. I also removed the model test, as it doesn't add much over the API test (although it respects the [Test Pyramid](https://martinfowler.com/bliki/TestPyramid.html))

@lwm how does this look to you?

Closes https://github.com/yunity/karrot-frontend/issues/1157


foodsaving/pickups/models.py
```diff
@@ -288,7 +288,7 @@ class PickupDateCollector(BaseModel):
     class Meta:
         db_table = 'pickups_pickupdate_collectors'
         unique_together = (('pickupdate', 'user'), )
-        ordering = ['-created_at']
+        ordering = ['created_at']
```
lwm: No migration needed?

tiltec: Ah right. It does create a migration, so I better commit it. But it doesn't have any effect on the database, so it works without.

foodsaving/conversations/api.py
```diff
@@ -139,7 +139,7 @@ def list(self, request, *args, **kwargs):
         pickup_conversations = [item for item in conversations if item.target_type == pickup_ct]
         pickups = PickupDate.objects. \
             filter(id__in=[c.target_id for c in pickup_conversations]). \
-            prefetch_related('collectors', 'feedback_given_by')
+            prefetch_related('pickupdatecollector_set', 'feedback_given_by')
```
lwm: :clap: :clap: :clap:

lwm:
LGTM :+1:

PR #605, opened 2019-01-06
Allow group members to view applicant user profiles.

lwm:
So, from my reading of https://github.com/yunity/karrot-frontend/issues/1117, allowing the group members to view the applicant profile resolves this.

Fixes https://github.com/yunity/karrot-frontend/issues/1117.

### DIFF
--- a/foodsaving/users/api.py
+++ b/foodsaving/users/api.py
@@ -46,8 +46,18 @@ class UserViewSet(mixins.RetrieveModelMixin, mixins.ListModelMixin, RetrievePriv
         return self.retrieve_private_conversation(request, pk)
 
     def get_queryset(self):
-        return self.queryset.filter(Q(groups__in=self.request.user.groups.all()) |
-                                    Q(id=self.request.user.id)).distinct()
+        is_member_of_group = Q(groups__in=self.request.user.groups.all())
+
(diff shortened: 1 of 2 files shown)


foodsaving/users/api.py
```diff
@@ -46,8 +46,18 @@ def conversation(self, request, pk=None):
...
 
     def get_queryset(self):
-        return self.queryset.filter(Q(groups__in=self.request.user.groups.all()) |
-                                    Q(id=self.request.user.id)).distinct()
+        is_member_of_group = Q(groups__in=self.request.user.groups.all())
+
+        is_self = Q(id=self.request.user.id)
+
+        group_ids = self.request.user.groups.values_list('id', flat=True)
+        is_applicant_of_group = Q(groupapplication__group__id__in=group_ids)
```
tiltec: Can you also write this instead?

```
groups = self.request.user.groups.all()
is_applicant_of_group = Q(groupapplication__group__in=groups)
```

From my understanding, it should produce the same SQL with less verbose Python.
But could be that you tried and it didn't work for some reason?

lwm: Ah, yes, that looks better! I will give it a shot.

PR #611, opened 2019-01-08
Change Conversation API from "email_notification" to "muted"

tiltec:
Now is more aligned with the conversation thread API. Needs frontend changes, which I will implement next.

---
Includes quite some refactoring to extend but simplify the ConversationSerializer. Using ConversationParticipant as base object instead of conversation has some benefits.
- no need to retrieve the participant from the request.user
- all conversation fields are read_only, can simply be retrieved with a 'source' kwarg
- participant fields such as seen_up_to can easily be read and written

On the flipside, the ConversationViewset seems a bit more complex, but only slightly.

_Could be better to use keep using the Conversation queryset for the viewset, then look up the participant and pass it to the serializer._


lwm:
A bit much that I couldn't really review it but is it related to https://github.com/yunity/karrot-frontend/issues/1174? Nice work :)

tiltec:
No, #1174 is untouched by this PR.

PR #588, opened 2018-11-30
Update emails.py

Ahrkano:
in calculate_group_summary_dates():
relativedelta(days=7) -> relativedelta(days=6)

PR #614, opened 2019-01-09
WIP: Weekly summary emails date fix

lwm:
@nicksellen, I've tried to add a test for https://github.com/yunity/karrot-frontend/issues/1083 that proves the thing is doing something wrong and am instead getting, what appears to me, the correct result (test is passing). If you could take a look at this, it'd be great.

My test sets up your example in https://github.com/yunity/karrot-frontend/issues/1083#issue-351889961 whereby the summary email is sent on Sunday the 19th (like the crontab specifies) and should give us the summary for 12th -> 18th. Am I reading that right?

Time remains confusing.

Something for https://github.com/yunity/karrot-frontend/issues/1083.
Refs: https://github.com/yunity/karrot-backend/pull/588

### DIFF
--- a/foodsaving/groups/tests/test_tasks.py
+++ b/foodsaving/groups/tests/test_tasks.py
@@ -1,3 +1,4 @@
+import datetime
 from datetime import timedelta
 from unittest.mock import patch
 
@@ -11,6 +12,7 @@ from config import settings
 from foodsaving.groups.factories import GroupFactory, PlaygroundGroupFactory, InactiveGroupFactory
 from foodsaving.groups.models import GroupMembership, GroupStatus
 from foodsaving.groups.tasks import process_inactive_users, send_summary_emails, mark_inactive_groups
+from foodsaving.groups.emails import calculate_group_summary_dates, prepare_group_summary_data
 from foodsaving.pickups.factories import PickupDateFactory, FeedbackFactory
 from foodsaving.stores.factories import StoreFactory
 from foodsaving.users.factories import UserFactory, VerifiedUserFactory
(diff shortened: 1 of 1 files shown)


nicksellen:
Ah, interesting. It definitely still sends Saturday -> Saturday emails as I get them each week. I'll have a peek.

Thanks for looking into it :)

lwm:
> Thanks for looking into it :)

Sure thing! I'll try to cause the bug when I look at this tomorrow.

nicksellen:
Aha, I found the cause, it's because of how it's rendered in the template:

https://github.com/yunity/karrot-backend/blob/c20446e2c549d1797ade28f05196ed123ae2c86a/foodsaving/groups/templates/group_summary.mjml#L12

Which uses the jinja2 date filter, which is defined as:

https://github.com/yunity/karrot-backend/blob/c20446e2c549d1797ade28f05196ed123ae2c86a/foodsaving/utils/email_utils.py#L19-L24

Which uses `get_current_timezone()` but there is no timezone override specified.

It could be added inside the task:

https://github.com/yunity/karrot-backend/blob/c20446e2c549d1797ade28f05196ed123ae2c86a/foodsaving/groups/tasks.py#L104

... or further down in the rendering.

nicksellen:
Do you want to implement the fix @lwm? Or I could.

lwm:
Aha! OK, I want to get back on this. I'll make time tomorrow, thanks for taking a look!

lwm:
Ack, I'm booked out this month and I don't think I'll get around to it any time soon :bomb:

PR #618, opened 2019-01-10
Add unsubscribe endpoint

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/1174

It works like this:
- backend generates a signed token that contains
  - user id
  - optional group id and name
  - optional conversation id OR thread id
- it gets put in a URL and added to the email footer e.g. `karrot.world/#/unsubscribe/<token>`
- when the user clicks it, the frontend will decode the data and present the user with choices:
  - unsubscribe from just this thread|conversation
  - unsubscribe from all conversations for this group
- the frontend will  `POST /api/unsubscribe` with a JSON payload of `{"choice": <choice>}` where choice is one of:
  - group
  - conversation
  - thread
- the backend will then decode the token, validate it, and perform the requested action

All of this can be done without being signed in, and will be the same even if you are signed in.

### DIFF
--- a/foodsaving/conversations/models.py
+++ b/foodsaving/conversations/models.py
@@ -98,20 +98,9 @@ class Conversation(BaseModel, UpdatedAtMixin):
         return type
 
     def find_group(self):
-        if self.is_private:
+        if self.is_private or self.target_type_id is None:
             return None
-        if self.target_type_id is None:
-            return None
-
-        type = self.type()
-        if type == 'pickup':
-            return self.target.store.group
-        if type == 'application':
-            return self.target.group
-        if type == 'group':
-            return self.target
-
-        raise Exception('cannot determine group for conversation {}'.format(self.id))
+        return self.target.group
 
 
 class ConversationParticipant(BaseModel, UpdatedAtMixin):
--- a/foodsaving/conversations/models.py
+++ b/foodsaving/conversations/models.py
@@ -97,6 +97,22 @@ class Conversation(BaseModel, UpdatedAtMixin):
 
         return type
 
+    def find_group(self):
+        if self.is_private:
+            return None
+        if self.target_type_id is None:
+            return None
+
+        type = self.type()
+        if type == 'pickup':
+            return self.target.store.group
+        if type == 'application':
+            return self.target.group
+        if type == 'group':
+            return self.target
+
+        raise Exception('cannot detirmine group for conversation {}'.format(self.id))
+
 
 class ConversationParticipant(BaseModel, UpdatedAtMixin):
     """The join table between Conversation and User."""
--- a/foodsaving/conversations/models.py
+++ b/foodsaving/conversations/models.py
@@ -111,7 +111,7 @@ class Conversation(BaseModel, UpdatedAtMixin):
         if type == 'group':
             return self.target
 
-        raise Exception('cannot detirmine group for conversation {}'.format(self.id))
+        raise Exception('cannot determine group for conversation {}'.format(self.id))
 
 
 class ConversationParticipant(BaseModel, UpdatedAtMixin):
--- a/foodsaving/unsubscribe/api.py
+++ b/foodsaving/unsubscribe/api.py
@@ -1,0 +1,50 @@
+from django.contrib.auth import get_user_model
+from django.utils.translation import ugettext as _
+from rest_framework import views, status
+from rest_framework.exceptions import ValidationError
+from rest_framework.permissions import AllowAny
+from rest_framework.response import Response
+
+from foodsaving.conversations.models import ConversationMessage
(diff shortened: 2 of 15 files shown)


foodsaving/conversations/models.py
```diff
@@ -97,6 +97,22 @@ def type(self):
...
+
+        type = self.type()
+        if type == 'pickup':
+            return self.target.store.group
+        if type == 'application':
+            return self.target.group
+        if type == 'group':
+            return self.target
+
+        raise Exception('cannot detirmine group for conversation {}'.format(self.id))
```
tiltec: typo:
```suggestion
        raise Exception('cannot determine group for conversation {}'.format(self.id))
```

foodsaving/conversations/models.py
```diff
@@ -97,6 +97,22 @@ def type(self):
...
         return type
 
+    def find_group(self):
+        if self.is_private:
+            return None
+        if self.target_type_id is None:
+            return None
+
+        type = self.type()
+        if type == 'pickup':
```
tiltec: I once had something like this implemented the other way round:
Each object would specify a property that refers to the group, e.g. `pickup.group` would return the group. The group itself also has a `group` property.
Therefore this code could be simplified to:

```
return self.target.group
```

foodsaving/unsubscribe/api.py
```diff
@@ -0,0 +1,50 @@
...
+    @staticmethod
+    def post(request, token):
+        """
+        Receive unauthenticated but signed unsubscribe requests
+        These are the things people can click in emails regardless of whether they are logged in
+        """
+
+        # TODO: make it return a generic "invalid token" error in most cases
+
+        choice = request.data.get('choice', 'conversation')
```
tiltec: Why would it fall back to `conversation`?

foodsaving/unsubscribe/utils.py
```diff
@@ -0,0 +1,82 @@
...
+def unsubscribe_from_thread(user, thread):
+    ConversationThreadParticipant.objects.filter(
+        user=user,
+        thread=thread,
+    ).update(
+        muted=True,
+    )
+
+
+def unsubscribe_from_all_conversations_in_group(user, group):
```
tiltec: I imagine that we easily forget to update this when we add another conversation type...

foodsaving/unsubscribe/api.py
```diff
@@ -0,0 +1,50 @@
...
+        data = parse_token(token)
```
tiltec: Does this return `None` when the token is invalid?

foodsaving/unsubscribe/api.py
nicksellen: Because it's the least impactful thing you can do, and the first option in the list, and kind of the thing that the link in the email says it will do "Unsubscribe from this conversation".... but also could just make it mandatory to provide a choice.

nicksellen:
@alangecker also mentioned the nice idea to use the list unsubscribe header, here's how it looks for a discourse email:

```
List-Unsubscribe: <https://community.coops.tech/email/unsubscribe/somelongtokenhere>
```

... although in gmail I don't actually see any UI for triggering it, but might be worth adding for the other clients (or maybe it does display in gmail in some cases).

Here is the [discourse discussion](https://meta.discourse.org/t/improving-our-unsubscribe-functionality/41448/30) about it, and the [resulting commit](https://github.com/discourse/discourse/commit/852860de6688668d6abe566cb1f4953fd13e2f35).

tiltec:
Sparkpost adds a List-Unsubscribe header by itself, which feeds into their suppression list (hence https://github.com/yunity/karrot-frontend/issues/1134). Could be good to replace it with our own header!

foodsaving/conversations/models.py
tiltec: It's partially implemented, here's an example: https://github.com/yunity/karrot-backend/blob/e5c048925ca24b3e99de1a96d76c0932a50240ad/foodsaving/pickups/models.py#L234-L236

nicksellen:
Note: to get a frontend url to try it out can run the following in a shell:

```python
from foodsaving.utils.frontend_urls import unsubscribe_url                                       
u = User.objects.first()                                                                         
g = u.groups.first()                                                                             
unsubscribe_url(u, g, g.conversation) 
```

foodsaving/unsubscribe/utils.py
nicksellen: I implemented it using a generic approach now. In the case of adding a new thing that has a conversation, the thing needs to have a `group` attribute. It will error if it does not.

foodsaving/unsubscribe/api.py
nicksellen: Frontend arrived!

nicksellen: https://github.com/yunity/karrot-frontend/pull/1195

PR #622, opened 2019-01-13
Add end date to pickups

nicksellen:
As part of https://github.com/yunity/karrot-frontend/issues/709


nicksellen:
Ok, it works! Although still some more thinking about how to make working with these ranges a bit nicer. I just put a lot of `date_range(date, minutes=30)` everywhere....

And probably some missing tests to see how updating it to have a nice end date works, etc...

... and then also we realised that part of the use case of having end dates is so that you could have very looooooooooooooooooooooooooooooooong pickups where you can do the pickup at any point, but in that case the useful thing is to be able to mark it as _done_ part way through.

The idea would be to add a _done_ field to pickups, and make our existing mechanisms use that too to check for _doneness_ (as opposed to just checking if the date passed and it had collectors), but that should probably be for another PR.

So my goal to make this one mergable is to allow the existing frontend to work with _minimal_ changes (but not bother about setting custom end dates - always have them 30 minutes).

PR #638, opened 2019-01-28
Resend invitation email

mvellasco:
Frontend PR: https://github.com/yunity/karrot-frontend/pull/1246

Finished the backend part for the issue #837.


foodsaving/invitations/api.py
```diff
@@ -66,3 +67,30 @@ def accept(self, request, **kwargs):
...
+    permission_classes = (
+        IsAuthenticated,
+    )
+    lookup_field = 'pk'
+
+    def get_queryset(self):
+        return self.queryset.filter(expires_at__gte=timezone.now())
+
+    @action(detail=True, methods=["POST"])
+    def resend_invitation_email(self, request, **kwargs):
```
tiltec: Can you move this action into the `InvitationsViewSet` and rename it to `resend`? 

I think it's useful if both share the same queryset, then we don't need to replicate the filtering. You also don't need to use an extra serializer (see other comment).

foodsaving/invitations/serializers.py
```diff
@@ -17,7 +17,9 @@ class Meta:
             UniqueTogetherValidator(
                 queryset=Invitation.objects.filter(expires_at__gte=timezone.now()),
                 fields=('email', 'group'),
-                message=_('An invitation has already been sent to this e-mail address')
+                message=_(
```
tiltec: I would keep this error message as is, it's getting shown when _creating_ a new invitation, not when resending it.

foodsaving/invitations/serializers.py
```diff
@@ -45,3 +47,20 @@ class InvitationAcceptSerializer(serializers.Serializer):
...
         invitation.accept(self.context['request'].user)
         return invitation
+
+
+class InvitationResendEmailSerializer(serializers.Serializer):
+    from datetime import timedelta
+
+    created_at = serializers.DateTimeField()
+
+    def validate(self, attrs):
```
tiltec: Why you pass the date from frontend to backend? The backend already knows about it (`invitation.created_at`) and the frontend is generally untrustworthy. It would be better to remove this.

foodsaving/invitations/serializers.py
```diff
@@ -45,3 +47,20 @@ class InvitationAcceptSerializer(serializers.Serializer):
...
+        if not timezone.now() >= attrs['created_at'] + self.timedelta(hours=1):
```
tiltec: I'd move this into a permission class and specify it in the action as
```
permission_classes = (IsAuthenticated, CanResend)
```

There are some examples in `pickups/api.py`.

foodsaving/invitations/serializers.py
```diff
@@ -45,3 +47,20 @@ class InvitationAcceptSerializer(serializers.Serializer):
...
+            raise serializers.ValidationError(
+                _('An invitation has already been sent to this e-mail address, you can only resend after one hour!')
+            )
+        return attrs
+
+    def update(self, invitation, validated_data):
+        invitation.resend_invitation_email()
```
tiltec: This can be called directly in the api action method if the permission checks pass, the serializer is not needed then :)

foodsaving/invitations/serializers.py
mvellasco: Sure, you're right, I'll fix that!

foodsaving/invitations/api.py
mvellasco: Yeah, sounds better, I've tried this approach but I got a little bit lost with the permissions, but after seeing the examples, I think I can do it.

foodsaving/invitations/serializers.py
mvellasco: I thought this was the best approach since I would have to pass the data manually to the serializer, but now that the serializer is not needed, that won't be a problem anymore.

foodsaving/invitations/serializers.py
mvellasco: Alright, I saw the examples, and I'll pick this up!

foodsaving/invitations/api.py
```diff
@@ -40,6 +47,19 @@ def get_throttles(self):
...
+        """
+        Resend invitation email
+        """
+        self.check_permissions(request)
+        instance = self.get_object()
+        self.check_object_permissions(request, instance)
+
+        instance.resend_invitation_email()
+
+        return Response(instance.created_at)
```
tiltec: I would either
- return nothing as it's not being used in the frontend
- or return the serialized invitation object, to have same type of response as the "create" action

You can do the latter by accessing the pre-defined invitation serializer:
```py
serializer = self.get_serializer(instance)
```

foodsaving/invitations/serializers.py
```diff
@@ -17,7 +17,9 @@ class Meta:
             UniqueTogetherValidator(
                 queryset=Invitation.objects.filter(expires_at__gte=timezone.now()),
                 fields=('email', 'group'),
-                message=_('An invitation has already been sent to this e-mail address')
+                message=_(
+                    'An invitation has already been sent to this e-mail address!'
```
tiltec: Maybe remove the added newlines, then the diff is clean and nice :)

foodsaving/invitations/api.py
```diff
@@ -20,6 +20,13 @@ def has_object_permission(self, request, view, obj):
         return not obj.group.is_member(request.user)
 
 
+class CanResend(BasePermission):
+    from datetime import timedelta
```
tiltec: Can you move this import to the top of the file? We usually don't have local imports.

(And we usually use `relativedelta` so you could do the same for consistency, but `timedelta` does the same in this case.)

foodsaving/invitations/api.py
mvellasco: I think it's best to return nothing. I did that return by mistake, sorry.

foodsaving/invitations/serializers.py
mvellasco: I forgot to reverse that change, after changing the message, sorry about that.

foodsaving/invitations/api.py
mvellasco: Yes, sure! I just did the import this way because the module was only used by this specific class, so I thought it was best this way. I'll make the change, no problem.

tiltec:
Oh, I think this branch is quite outdated by now...
@mvellasco are you still interested in working on this branch?

mvellasco:
Yes @tiltec I'm still working on it, but this last month has been quite complicated for me, but I'll finish the work hopefully by the end of the week.

PR #642, opened 2019-01-29
Rename feedback_possible field to is_done

nicksellen:
This is part of https://github.com/yunity/karrot-frontend/issues/709 - with the idea that if pickups had a long duration, they might be done part way through, and so would be useful to be able to mark them as done.

This doesn't add the API to mark them as done. Yet.


nicksellen:
Rebase/renamed so it fits on top of master now.

PR #652, opened 2019-02-01
Add duration field to pickup series

nicksellen:
Pickup series need to have a duration specified now to create the correct date range.


foodsaving/pickups/models.py
```diff
@@ -41,7 +42,8 @@ class PickupDateSeries(BaseModel):
 
     def create_pickup(self, date):
         return self.pickup_dates.create(
-            date=CustomDateTimeTZRange(date, date + timedelta(minutes=30)),  # TODO: make duration part of series?
+            date=CustomDateTimeTZRange(date, date + (self.duration if self.duration else default_duration)),
```
nicksellen: `self.duration or default_duration`

PR #658, opened 2019-02-09
Use timezone when rendering emails

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/1083

I noticed we didn't really specify the timezone much when rendering emails - I added it as a parameter to the `prepare_email()` function now to encourage it to be passed through (defaults to utc) and eventually call `with timezone.override()` just as it renders the template.

The other option would have been to put the email preparing inside a `with timezone.override()` much higher up for each case, but found it cleaner to pass it in directly (and revealed it's not very clear which timezone should be for private messages, as they are outside the context of a group).

### DIFF
--- a/foodsaving/applications/emails.py
+++ b/foodsaving/applications/emails.py
@@ -22,6 +22,7 @@ def prepare_new_application_notification_email(user, application):
         template='new_application',
         from_email=from_email,
         user=user,
+        tz=application.group.timezone,
         reply_to=[reply_to],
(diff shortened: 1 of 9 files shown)


nicksellen:
I also added in a sort-of-vaguely-related thing to make it always send the summary emails at 8am on Sunday in _local_ time (currently uses UTC).

tiltec:
Ah, there's a test failure: python 3.5 doesn't support the `**kwargs` syntax when declaring a dict.

PR #663, opened 2019-02-15
Rename foodsaving to karrot

nicksellen:
As part of our generalization plans, it makes sense to rename our main "foodsaving" directory (which is really just a relic of when it was called the "foodsaving tool").


tiltec:
Looks good! I wonder if this breaks some assumptions in the ansible playbook, but I think it should be fine!
Any other areas to watch out for?

nicksellen:
> Looks good! I wonder if this breaks some assumptions in the ansible playbook, but I think it should be fine!

Ah yeah. I had a quick look, but all looked fine. Doesn't mention foodsaving other than in domains.

>  Any other areas to watch out for?

I don't think so... press merge?

PR #680, opened 2019-03-04
Added welcome_message to Group model and application_accepted email

cstefanj:
@tiltec Not ready yet but a first stab at [#1038](https://github.com/yunity/karrot-frontend/issues/1038). Opening a draft so we can move the conversation here.


cstefanj:
Moving the unrelated local test failure here from the issue thread:
[image](https://user-images.githubusercontent.com/1835675/53757165-3e81a100-3ebb-11e9-8d4f-de759a3f0cef.png)

cstefanj:
@tiltec Hi again. This works now, the updated email looks like this. I don't know why that one CI test fails, it passes on my local machine. Thoughts on what I should do next? :-) 
image

tiltec:
Looks good! I'll have a look at the code and the CI error soon.

karrot/applications/templates/application_accepted.mjml
```diff
@@ -9,6 +9,7 @@
             <mj-column>
                 <mj-text>
                     <h1>{% trans group_name=group.name %}Your application to join {{ group_name }} was accepted{% endtrans %}</h1>
+                    <p> {% trans group_msg=group.welcome_message %} {{ group_msg }} {% endtrans %}</p>
```
tiltec: This can't be translated, you can remove the `trans` tags and directly write `group.welcome_message` into the double curly braces.

tiltec:
The CI failure is definitely unrelated to your changes, I had problems with this test before that I didn't understand. It's probably a interdependency between this and other tests. You can just ignore it.

karrot/applications/templates/application_accepted.mjml
cstefanj: That makes perfect sense of course since this is a user input. Translating would require some trickery with the Google Translate API perhaps. :-) But I assume that the admin(s) will write the welcome message in the local language.

I have made the change now.

karrot/applications/templates/application_accepted.mjml
```diff
@@ -9,6 +9,7 @@
             <mj-column>
                 <mj-text>
                     <h1>{% trans group_name=group.name %}Your application to join {{ group_name }} was accepted{% endtrans %}</h1>
+                    <p> {{ group.welcome_message }} </p>
```
tiltec: May need to wrap this into a `if` block, as we don't want to have an empty `p` tag if the group didn't specify a welcome message.

cstefanj: Good point. `{% if group.welcome_message %}<p> {{ group.welcome_message }} </p>{% endif %}` should cover both when `welcome_message` is `False` and `""` if I'm not mistaken.

cstefanj: The case this doesn't cover would be if the field has only whitespace in it. Not sure if that's enough of a problem to handle - and if so, should it be handled here or as a check when you save the field?

tiltec: Good catch! I would ignore the "only whitespace" case for now, we also don't handle it for other fields and it doesn't seem to be a problem.

cstefanj: Sounds good. Anything else you want me to handle in this PR? I'm thinking about the invitation email, or joining an open group. Otherwise I think it should be ok to merge.

PR #692, opened 2019-03-31
add support for discourse email_reply_trimmer

tiltec:
Closes https://github.com/yunity/karrot-frontend/issues/1096

`email_reply_trimmer` works better than `talon` in almost all cases, for a few exceptions. In all cases, it seems safe to choose the trimmed reply that has less lines.

To enable, configure `EMAIL_REPLY_TRIMMER_URL` in `local_settings.py`.

If the service goes down, it should cause no error. Maybe we should send a warning to sentry though.

It still needs some changes to our ansible setup to deploy the docker container: https://github.com/yunity/email_reply_trimmer_service

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -280,6 +280,10 @@ HUEY = {
     'always_eager': True,
 }
 
+# If you have the email_reply_trimmer_service running, set this to 'http://localhost:4567/trim' (or similar)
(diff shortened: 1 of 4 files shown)


tiltec:
Should I run one instance of `email_reply_trimmer_service` on yuca for all Karrot instances? Or one per deployment?

PR #742, opened 2019-05-31
Add feedback export API

tiltec:
Related to https://github.com/yunity/karrot-frontend/issues/1455


tiltec:
Test fails because Python 3.5 doesn't support the `timespec` argument to `isoformat()` yet. Will remove...

PR #835, opened 2019-10-08
Use group image in email header

nicksellen:
Closes https://github.com/yunity/karrot-frontend/issues/1770


tiltec:
Do you think it's ok if the picture breaks sometimes? I imagine the case where the group changes their image and the backend garbage collects the old image. Maybe that's fine, as most users won't look at their old notification emails anyways.

nicksellen:
I guess the fancy way would be to implement a special image fixed URL that gives you the current one,..e.g /groups/1/logo.jpg.

nicksellen:
I just need to add a test to test my changes too...

karrot/utils/email_utils.py
```diff
@@ -96,6 +97,12 @@ def prepare_email(
         raise Exception('email context should not redefine defaults: ' + ', '.join(redefined_keys))
     context.update(default_context)
 
+    if 'header_image' not in context:
+        if 'group' in context and context['group'].photo:
+            context['header_image'] = context['group'].photo.url
+        else:
+            context['header_image'] = settings.DEFAULT_EMAIL_HEADER_IMAGE
```
tiltec: This seems to result in `{HOSTNAME}https://user-images.githubusercontent.com/31616/36565633-517373a4-1821-11e8-9948-5bf6887c667e.png`, at least I didn't notice anything that should prevent it.

nicksellen: Ah yeah, of course. Doh!

nicksellen: Should be fixed now. I'm trying to learn how to program again...

PR #855, opened 2019-11-03
Add offers API

nicksellen:
Backend implemention for offers.

See https://community.foodsaving.world/t/281 and https://github.com/yunity/karrot-frontend/pull/1799


nicksellen:
I added a slightl funky way to be able to do POST/PATCH requests with bodies containing JSON and multiple images at the same time... as the offers allow multiple images and I don't think it would be very good UI to have separate image handling as for profile/group images.

The more normal way would probably be to upload the images before/after the main data, but I wanted to try doing it all in one multipart request.

karrot/history/migrations/0010_auto_20191103_1957.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.6 on 2019-11-03 19:57
+
+from django.db import migrations
+import django_enumfield.db.fields
+import karrot.history.models
+
+
+class Migration(migrations.Migration):
```
nicksellen: I shouldn't have committed this file.

tiltec: Resolved with https://github.com/yunity/karrot-backend/commit/cb8a4dd52f69bcb399ab8bf8991a202c17e5f3e0

tiltec:
> I added a slightly funky way to be able to do POST/PATCH requests with bodies containing JSON and multiple images at the same time... as the offers allow multiple images and I don't think it would be very good UI to have separate image handling as for profile/group images.

Great idea! Shouldn't be too hard to unit test :)

nicksellen:
So, should be good for review now! @tiltec 

It's worth trying it out with the work-in-progress frontend (at https://github.com/yunity/karrot-frontend/pull/1799 - branch is "sharing"), the frontend is not polished/finished yet, but most of the features are sort of working...

nicksellen:
... one bit I haven't really tried is the conversation that goes with it - the frontend currently has a mock one displayed. I sort of assumed that will all JustWork, but maybe not.

nicksellen:
... also, I refer to the offer pictures as "image" as opposed to "photo" (used for group/user). This seems inconsistent. I did it because they might not be a _photo_, could be a drawing... but would prefer consistency, so some options:
1. leave it as-is
2. rename offer "images" to "photos"
3. rename all other "photo" things to "image"

I would be inclined to do 2, and consider doing 3 at some unspecified point in the future if desired.

tiltec:
> I would be inclined to do 2, and consider doing 3 at some unspecified point in the future if desired.

I'm happy either way, 2 seems fine! I agree that the "photo" term is not always appropriate.

tiltec:
I didn't try the conversation stuff either, usually you might encounter some problems while sending message notification emails if the configuration is not complete (e.g. there's no sending function for offer message emails yet)

karrot/offers/api.py
```diff
@@ -0,0 +1,146 @@
...
+    filterset_fields = (
+        'group',
+        'status',
+    )
+    pagination_class = OfferPagination
+    permission_classes = (
+        IsAuthenticated,
+        IsOfferUser,
+    )
+    parser_classes = [JSONWithFilesMultiPartParser, JSONParser]
```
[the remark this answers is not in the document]
nicksellen: It means you are not _forced_ to use the multipart format, useful if you are just editing a normal field. It get used in `test_update_offer` and `test_reposition_image` offer tests.

karrot/offers/api.py
```diff
@@ -0,0 +1,146 @@
...
+    @action(
+        detail=True,
+        methods=['GET'],
+        permission_classes=(),
+    )
+    def image(self, request, pk=None):
+        image = OfferImage.objects.filter(offer=pk).first()
+        if not image:
+            raise Http404()
+        return HttpResponseRedirect(redirect_to=image.image.url)
```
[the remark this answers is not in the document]
nicksellen: vue-croppa is resizing images before upload, I just chose a 10mb image for my dev.karrot.world profile, and the "fullsize" image ends up as https://dev.karrot.world/media/user__photos/photo_cntAK1q.jpeg (~212kb), and the request meta in firefox confirms that upload content length is 217423.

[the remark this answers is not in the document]
nicksellen: ... technically they could upload it outside of karrot UI and get a bigger photo I guess ...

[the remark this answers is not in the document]
tiltec: vue-croppa usually works fine, but I sometimes encountered really big images around 1MB (I think it was PNG files instead of JPEG). And the API allows all image formats and sizes anyways. But as this problem already exists with the other image usages (group photo and user profile picture), you don't have to solve it for the offers. Just to keep in mind that it _could_ become problematic to default to the fullsize version.

nicksellen:
Ok, did a few more bits:

> Maybe add a push notification for new offers? But that might need an option to disable the push notification.

I added it _blind_ (didn't try it...). I used the group notification setting for NEW_OFFERS to decide whether to send it.

> Maybe let the conversation close after the offer is accepted or disabled? There's the ended_at property for that, you might need another updated_at field on the Offer model for that though: 

Done :) I actually switched to the explicit accept/archive actions rather than just updating the status field (and also changed "disabled" to "archived").

> I didn't try the conversation stuff either, usually you might encounter some problems while sending message notification emails if the configuration is not complete (e.g. there's no sending function for offer message emails yet)

I added the sending functions, and DRY'd it up a bit. Hopefully it all works.

karrot/subscriptions/tasks.py
```diff
@@ -119,14 +121,63 @@ def notify_message_push_subscribers_with_language(message, subscriptions, langua
...
+    )
+
+
+@db_task()
+def notify_new_offer_push_subscribers(offer):
+
+    users = offer.group.members.filter(
+        groupmembership__in=GroupMembership.objects.active().with_notification_type(GroupNotificationType.NEW_OFFER),
+    ).exclude(
+        id__in=get_user_model().objects.unverified_or_ignored(),
```
tiltec: This is just relevant for emails, I think we can still push to users with unverified or ignored email addresses.

karrot/subscriptions/tasks.py
```diff
@@ -119,14 +121,63 @@ def notify_message_push_subscribers_with_language(message, subscriptions, langua
...
+
+    notify_subscribers_by_device(
+        subscriptions,
+        click_action=frontend_urls.offer_url(offer),
+        fcm_options={
+            'message_title': message_title,
+            'message_body': Truncator(offer.description).chars(num=1000),
+            # this causes each notification for a given conversation to replace previous notifications
+            # fancier would be to make the new notifications show a summary not just the latest message
+            'tag': 'conversation:{}'.format(offer.id),
```
tiltec: Change to 'offer:{}'

karrot/subscriptions/tasks.py
```diff
@@ -119,14 +121,63 @@ def notify_message_push_subscribers_with_language(message, subscriptions, langua
     else:
         click_action = frontend_urls.conversation_url(conversation, message.author)
```
tiltec: `conversation_url` does not handle offer conversations yet

tiltec:
> I used the group notification setting for NEW_OFFERS to decide whether to send it.

Group notification settings are explicitly marked as "email notifications" in the frontend. Just to keep in mind for https://github.com/yunity/karrot-frontend/issues/1457

Quite a lot of refactoring in this branch! I'd propose a strategy of "release it while it's still hot" if you're around to iron out the bugs then :)

nicksellen:
Thanks for further review, all addressed now. Yes, I would be up for "release whilst it's hot". I'm not going anywhere.

PR #878, opened 2019-12-02
Create offer transactionally

nicksellen:
We create offers with their images. But using our normal signals method to trigger the notifications, they would get triggered without their associated images available (see https://sentry.io/organizations/foodsaving-worldwide/issues/1345296602/).

This PR makes it so we create the offer and it's images inside a transaction. That doesn't entirely solve it though as the post_save signal is still triggered during the transaction. But we can use the `transaction.on_commit` hook if we are in a transaction to only trigger the notification when the transaction is closed.

If we had to do this is many places it might be nice to wrap it up into an abstraction, but at least for now this is the only model that has dependent models like that.

### DIFF
--- a/karrot/offers/receivers.py
+++ b/karrot/offers/receivers.py
@@ -14,12 +14,6 @@ def offer_saved(sender, instance, created, **kwargs):
         conversation = Conversation.objects.get_or_create_for_target(offer)
         conversation.join(offer.user)
 
-        def notify():
-            notify_members_about_new_offer(offer)
-
         # offer saving is normally done in a transaction so as to include the images
         # we only want to trigger the notification after this transaction is complete
-        if transaction.get_connection().in_atomic_block:
-            transaction.on_commit(notify)
-        else:
-            notify()
+        transaction.on_commit(lambda: notify_members_about_new_offer(offer))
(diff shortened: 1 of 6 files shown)


karrot/offers/serializers.py
```diff
@@ -61,9 +62,12 @@ def save(self, **kwargs):
 
     def create(self, validated_data):
         images = validated_data.pop('images')
-        offer = Offer.objects.create(**validated_data)
-        for image in images:
-            OfferImage.objects.create(offer=offer, **image)
+        # Save the offer and it's associated images in one transaction
```
tiltec:
```suggestion
        # Save the offer and its associated images in one transaction
```

karrot/offers/receivers.py
```diff
@@ -13,4 +14,12 @@ def offer_saved(sender, instance, created, **kwargs):
         conversation = Conversation.objects.get_or_create_for_target(offer)
         conversation.join(offer.user)
 
-        notify_members_about_new_offer(offer)
+        def notify():
+            notify_members_about_new_offer(offer)
+
+        # offer saving is normally done in a transaction so as to include the images
+        # we only want to trigger the notification after this transaction is complete
+        if transaction.get_connection().in_atomic_block:
```
tiltec: Docs say: 

> If you call on_commit() while there isn’t an active transaction, the callback will be executed immediately.

Did you try it without the check?

nicksellen: Yup, good catch, I removed the check (I confirmed it still ran when the receiver wasn't called in a transaction, but also just read the code, it's quite clear).

PR #882, opened 2019-12-03
Add /api/stats/ endpoint

nicksellen:
The backend for https://github.com/yunity/karrot-frontend/pull/1916

It:
- accepts JSON post requests to `/api/stats`
- which can contain a list of stats (up to 50 points in one request)
- it's throttled to 60/minute
- it is accessible without auth (as the frontend also records logged out views) (hence the limits/throttling to prevent someone filling up our influxdb too fast...)
- records the stats into `karrot.stats.frontend`
- puts `ms` as a field, and the rest as tags


nicksellen:
... I could also make it just accept any additional fields to use as tags... but thought maybe nice to keep it restricted...

PR #887, opened 2019-12-04
Fixes #1769 Invitation emails are always in English

lenniezelk:


karrot/invitations/tests/test_prepare_email.py
```diff
@@ -0,0 +1,32 @@
...
+from karrot.invitations.models import Invitation
+from karrot.invitations.emails import prepare_emailinvitation_email
+from karrot.utils.frontend_urls import invite_url
+
+
+class PrepareEmailTestCase(TestCase):
+    def test_prepare_email_called_with_language(self):
+        invitation: Invitation = InvitationFactory(invited_by__language='sw')
+
+        with patch('karrot.invitations.emails.prepare_email') as pe:
```
nicksellen: We tend to use the @patch version of it, less indention needed :) An example:

https://github.com/yunity/karrot-backend/blob/cfbcdb905e7ab6a01d5d5c60c21d1f1e8da53abf/karrot/groups/tests/test_receivers.py#L72-L75

karrot/invitations/tests/test_prepare_email.py
```diff
@@ -0,0 +1,32 @@
...
+            prepare_emailinvitation_email(invitation)
+            pe.assert_called_once_with(
```
nicksellen: Would be better to just check the `language`, otherwise it'll start failing if we make unrelated changes to the `context` (for example).

You can get the kwargs out with something like:
```py
args, kwargs = pe.call_args
```
Then use the normal `self.assert*` methods to check the language.

karrot/invitations/tests/test_prepare_email.py
nicksellen: Also, I'd named it `prepare_email` for clarity.

lenniezelk:
updated @nicksellen

karrot/invitations/tests/test_api.py
```diff
@@ -164,6 +164,13 @@ def test_invite_as_newcomer_fails(self):
         response = self.client.post(base_url, {'email': 'please@join.com', 'group': self.group2.id})
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
+    def test_invite_with_different_invited_by_language(self):
+        self.member.language = 'fr'
+        self.member.save()
+        self.client.force_login(self.member)
+        response = self.client.post(base_url, {'email': 'please@join.com', 'group': self.group.id})
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
tiltec: @lenniezelk 
Could you add an assertion that the email is actually in Swedish (or not in English, as that would be the default)?
You should be able to access the subject via `mail.outbox[0].subject`, for example.

Otherwise looks great!

lenniezelk:
the translation doesn't seem to be working. I have added some logs to see why. The correct language is indeed passed to the email util function here `karrot/utils/email_utils.py::prepare_email_content` and in the `with timezone.override(tz), translation.override(language):` block `translation.get_language()` returns correct language code but the subject is not translated. Any ideas what is happening? @tiltec @nicksellen

nicksellen:
Having a look... I think the locale is not activated in the tests or something. At first I noticed that `test_settings.py` is missing `django.middleware.locale.LocaleMiddleware`, but after adding it, it still didn't work as expected. Digging more...

tiltec:
Did you try a language where the subject is actually translated? I believe Swedish might be lagging behind.

nicksellen:
This test is for French, and the translation is available:

```
#: karrot/invitations/templates/emailinvitation.subject.jinja2:1
#, python-format
msgid "Invitation to join %(group_name)s"
msgstr "Invitation de joinde %(group_name)s"
```

Whereas in the test I get:
```
[karrot local development] Invitation to join Group 0 Stephanie Smith
```

Language is set correctly (I added a function to the jinja2 environment to print it out inside the template...).

So I think it's something about how the test environment is setup...

nicksellen:
... so the translation is activated, but it doesn't contain our translations, only some other django-y ones, e.g. `This field cannot be null.` (will be correctly translated).

I added logging to where it adds in the various locales, and can see it adding karrot translations though... dig dig dig ...

nicksellen:
... the tests are only reading the `.mo` (compiled) translation messages, not the `.po` (uncompiled) ones ... (so it works after doing `./manage.py compilemessages`) ... but maybe it can be setup to use the uncompiled ones ...

nicksellen:
... ok, so I think there is no support for _on the fly_ `.po` file compilation, and we just have to have the messages compiled... maybe we should actually commit the compiled messages to the git repo too... is there any advantage to not having them there (it also affects all translations in development)?

As an aside, I also noticed we are not really setting the language properly, so if there is a validation error, that's also just in english (it will be translated if there are `.mo` files _and_ I set the `Accept-Language` header), looks like we can do that with the session `LANGUAGE_SESSION_KEY` (or a cookie `LANGUAGE_COOKIE_NAME`). But that's not for this PR I think.

tiltec:
Right! Yeah, I think we should commit `.mo` files - also [django itself](https://github.com/django/django/tree/master/django/conf/locale/be/LC_MESSAGES) is doing so.
We might still need to run `compilemessages` on deployment though, in case a dependency doesn't ship `.mo` files.

> As an aside, I also noticed we are not really setting the language properly, so if there is a validation error, that's also just in english (it will be translated if there are .mo files and I set the Accept-Language header),

We _do_ set the `Accept-Language` header for frontend requests, so that should be fine.

nicksellen:
Aha I see `Accept-Language` in the frontend now, I had expected the backend to check, given we have the logged in user, but yeah, seems fine... I was trying it with fetch API, so it didn't have the axios headers.

So! @lenniezelk you could run `./manage.py compilemessages` and add the `*.mo` files to the git repo (need to remove the `*.mo` entry in `.gitignore too`), then it should all work!

tiltec:
> Aha I see Accept-Language in the frontend now, I had expected the backend to check, given we have the logged in user, but yeah, seems fine...

That's mostly for historic reasons - before we sent out emails, we didn't need to keep user's language choice in the backend, so `Accept-Language` was the way to go.
I'd keep using `Accept-Language`, as it also works for logged-out users, e.g. for validation errors and emails during signup and login.

lenniezelk:
updated. Thanks for the help. @nicksellen @tiltec

PR #888, opened 2019-12-04
Avoid high number of values per tag in influxdb

tiltec:
Background: https://github.com/bitlabstudio/django-influxdb-metrics/issues/27

route_path may contain many different values because it contains many IDs, so we shouldn't keep it as tag.

route_params may also grow in size (especially with something like so pickup_ids). We should keep an eye on it.

### DIFF
--- a/karrot/stats/serializers.py
+++ b/karrot/stats/serializers.py
@@ -15,7 +15,6 @@ class StatsEntrySerializer(serializers.Serializer):
     group = serializers.IntegerField(allow_null=True)
     route_name = serializers.CharField()
     route_path = serializers.CharField()
-    route_params = serializers.DictField(allow_empty=True)
 
     # device/build info
     mobile = serializers.BooleanField()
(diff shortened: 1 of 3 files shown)


nicksellen:
Yes, good point! Maybe we don't even need the route params at all? (given we've got the group?)

tiltec:
Agree, having the route params should be sufficient!

Actually, in addition to that, we could also lift the 'max-values-per-tag' limit in influxdb: https://docs.influxdata.com/influxdb/v1.7/administration/config/#max-values-per-tag-100000
Not sure what the implications would be, but it seems other people do this just fine: https://stackoverflow.com/questions/43770354/max-values-per-tag-limit-exceeded-influxdb

nicksellen:
> Agree, having the route params should be sufficient!

Does not agree with the sentence you replied to: _Yes, good point! Maybe we don't even need the route params at all? (given we've got the group?)_

What did you mean?

tiltec:
Indeed, I misread!
I would like to have a way to identify the page the user is on. Biggest difference is probably in user-generated content. I think we need either the params or the full path, and both seems fine to have as fields. (InfluxDB can do regex queries in fields, which I found sufficient for some manual drill-down analysis)

Given the full path seems easier to understand, let's remove the route params.

nicksellen:
Yup, seems fine to me. Keep the group though!

I think for my interest, the overall figures might be most interesting.

PR #958, opened 2020-05-01
made default group type/theme customizeable by instance

inktrap:
I also added a manager for groups, which is compatible to the previous approach. It may be helpful during the cleaner implementation of group types.

- [x] review
- [x] run tests (with different configs)
- [x] decide if the group manager is useful
- [x] (perhaps) test manager

I am not sure how we should test the new manager, since it is compatible to the previous approach. Coverage would thank us though.

Since I changed settings.py, it is important to check that all the different settings files still work. So, please review/test this to make sure.


inktrap:
Since I changed settings.py, it is important to check that all the different settings files still work. So, please review/test this to make sure.

tiltec:
I think the file should be called `groups/themes.py` instead of `groups/grouptypes.py` - sorry for my misleading post yesterday!

karrot/groups/models.py
```diff
@@ -62,9 +54,12 @@ def annotate_yesterdays_member_count(self):
             )
         )
 
+# to avoid circular imports, managers has to be imported after GroupQuerySet
+from karrot.groups import managers
 
 class Group(BaseModel, LocationModel, ConversationMixin, DirtyFieldsMixin):
-    objects = GroupQuerySet.as_manager()
+    # objects = GroupQuerySet.as_manager()
+    objects = managers.GroupManager()
```
tiltec, line 62: Have a look at https://docs.djangoproject.com/en/3.0/topics/db/managers/#from-queryset to avoid having to duplicate queryset methods in the custom manager.

inktrap:
I don't like the name grouptypes either. It also contains status info.
I think it is really important to distinguish group types and themes conceptually! In karrot this seems pretty identical/coupled right new.

tiltec:
It's called group themes because it only causes visual changes - functionality is always the same.

karrot/groups/models.py
inktrap: The changes I made in my master branch should show up I here automatically, right?
I have currently no time to test this (because I am going to paint our fairteiler now.)

inktrap:
But it feels like more than a theme, because error messages of the backend, mail templates (basicly all the stuff from the other issues about the bikekitchen theme) have to be adapted too …

Just for clarification: Groups should be able to change their background without changing the group theme, right? (karrot-dev is probably a better place to discuss this, but I have to leave now, sorry.)

config/settings.py
```diff
@@ -44,6 +44,10 @@
 # For removing inactive users from groups
 NUMBER_OF_INACTIVE_MONTHS_UNTIL_REMOVAL_FROM_GROUP_NOTIFICATION = 6
 NUMBER_OF_DAYS_AFTER_REMOVAL_NOTIFICATION_WE_ACTUALLY_REMOVE_THEM = 7
+# Set the default theme for new groups: GENERAL, FOODSAVING, BIKEKITCHEN
+GROUP_THEME_DEFAULT = grouptypes.GroupTheme.FOODSAVING
+# Set the default status for new groups, ACTIVE, PLAYGROUND, INACTIVE
+GROUP_STATUS_DEFAULT = grouptypes.GroupStatus.ACTIVE
```
tiltec, line 50: Why did you move that as well?

karrot/groups/models.py
```diff
@@ -62,9 +54,11 @@ def annotate_yesterdays_member_count(self):
             )
         )
 
+# to avoid circular imports, managers has to be imported after GroupQuerySet
```
tiltec, line 57: For consistency with other modules you should move the manager in here, right above or below GroupQuerySet (have a look at other custom managers in Karrot)

config/settings.py
```diff
@@ -44,6 +44,10 @@
 # For removing inactive users from groups
 NUMBER_OF_INACTIVE_MONTHS_UNTIL_REMOVAL_FROM_GROUP_NOTIFICATION = 6
 NUMBER_OF_DAYS_AFTER_REMOVAL_NOTIFICATION_WE_ACTUALLY_REMOVE_THEM = 7
+# Set the default theme for new groups: GENERAL, FOODSAVING, BIKEKITCHEN
```
tiltec, line 47: Better refer to the enum instead of duplicating the values here.

config/settings.py
inktrap: I didn't move it, I want to guarantee that the variables are set and have a default value that can be overwritten by local_settings.py and that they are not hardcoded in the manager or in models.py (however, it might be better to have them in models.py …)

inktrap: The problem with that then is that they *have* to be present in local_settings.py … so I wanted to guarantee that they are present by adding them to settings.py

inktrap:
Cleaned it up and tested it, also fixed flake8 warnings.

config/settings.py
tiltec: I mean why you made the default GroupStatus into a setting?

karrot/groups/models.py
```diff
@@ -63,8 +57,21 @@ def annotate_yesterdays_member_count(self):
         )
 
 
+class GroupManager(models.Manager):
+    def create(self, *args, **kwargs):
+        try:
+            kwargs['theme'] = kwargs.get('theme', settings.GROUP_THEME_DEFAULT.value)
```
tiltec, line 63: That was a misunderstanding, it's definitely better to always have it present in settings.py instead of catching AttributeError.

inktrap: Okay. I changed it again.

The reason why I don't want to write something like

```
default=themes.GroupTheme.FOODSAVING
```

in ``models.py`` is, because it should be immediately clear which default value is actually used (settings.py or models.py), so everything, including the manager in models.py refers back to settings explicitly.

karrot/groups/models.py
```diff
@@ -63,8 +57,15 @@ def annotate_yesterdays_member_count(self):
         )
 
 
+class GroupManager(models.Manager):
```
tiltec, line 60: I just noticed the _Karrot way_ seems to be to inherit from `BaseManager.from_queryset(GroupQuerySet)`. Then you can directly specify `objects = GroupManager()` in the model. Have a look at `UserManager`, `ConversationMessageManager` or `NotificationManager` for reference.

karrot/groups/models.py
```diff
@@ -63,8 +57,15 @@ def annotate_yesterdays_member_count(self):
...
+    def create(self, *args, **kwargs):
+        kwargs['theme'] = kwargs.get('theme', settings.GROUP_THEME_DEFAULT.value)
+        kwargs['status'] = kwargs.get('status', settings.GROUP_STATUS_DEFAULT.value)
```
tiltec, line 63: I think this can be removed, or do you need to override the default group status? I commented on this before, but I didn't understand your rationale so far.

inktrap: It would be useful to set up a playground in an automated way. Like dev.foodsaving.world there could be a dev.plantsharing.de. Don't you think it would be useful to make karrot more configureable?

However, it's not nice that ``themes`` does not fit well if the status-enum is included in there …

karrot/groups/models.py
```diff
@@ -78,13 +79,15 @@ class Group(BaseModel, LocationModel, ConversationMixin, DirtyFieldsMixin):
...
+        # default can be overwritten by local_settings.py GroupManager
+        default=settings.GROUP_STATUS_DEFAULT.value,
+        choices=[(status.value, status.value) for status in themes.GroupStatus],
         max_length=100,
     )
     theme = models.TextField(
-        default=GroupTheme.FOODSAVING.value,
-        choices=[(theme.value, theme.value) for theme in GroupTheme],
+        # default can be overwritten by local_settings.py GroupManager
+        default=settings.GROUP_THEME_DEFAULT.value,
```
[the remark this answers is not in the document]
inktrap: Why do you want to remove choices? It would be necessary if we'll create a dialog so people can actually create differently themed groups. (see bike kitchen issues)

karrot/groups/models.py
inktrap: Alright!

inktrap: Is it okay if I mark stuff as resolved or do you want to do it?

karrot/groups/models.py
inktrap: (Your previous comment must have slipped by, sorry)

inktrap:
> A test would also be nice :)
> https://docs.djangoproject.com/en/3.0/topics/testing/tools/#django.test.override_settings might be helpful for this.

I implemented a test for group creation, didn't know how to test the manager directly. The goal can't be to test django or the db.

Also I added a small section in README.md about running tests selectively.

karrot/groups/models.py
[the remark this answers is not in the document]
tiltec: I mean to remove `default=settings.GROUP_THEME_DEFAULT.value`, as this default is already provided by the new manager.

karrot/groups/models.py
tiltec: > It would be useful to set up a playground in an automated way. Like dev.foodsaving.world there could be a dev.plantsharing.de.

Maybe we should talk about that in a different issue or PR, I'm not sure if this really helps in the way you intend it.

> Don't you think it would be useful to make karrot more configureable

Yes, on a case-by-case basis :)

README.md
```diff
@@ -56,14 +56,28 @@ docker-compose exec backend ./manage.py migrate
...
 
+*Run tests selectively:* If you want to run only a single test, let's say
+`TestGroupManager` in `karrot/groups/tests/test_model.py`, you can do so by
+using dot-syntax:
+
+```
+docker-compose exec backend ./manage.py test karrot.groups.tests.test_model.TestGroupManager
+```
+
+Tipp: To speed up test development, ``pprint`` might be useful to you.
```
tiltec, line 78: Maybe "Hint"? https://dict.leo.org/englisch-deutsch/tipp

karrot/groups/tests/test_model.py
```diff
@@ -81,3 +82,50 @@ def test_does_not_count_from_other_groups(self):
...
+
+
+class TestGroupManager(TestCase):
+
+    # test if setting a default status and theme via settings works
+
+    # set each setting individually
+    @override_settings(GROUP_THEME_DEFAULT=themes.GroupTheme.FOODSAVING)
+    def test_create_group_theme_default_foodsaving(self):
+        this_object = Group.objects.create()
```
tiltec, line 94: Looks good!
Maybe write `group = GroupFactory()` for consistency with other tests (have look at `TestGroupModel` in the same file).

inktrap:
https://github.com/yunity/karrot-backend/pull/958#discussion_r420127525

What about the status? Should I remove that? We can talk about this in #karrot-dev. :)

tiltec:
> What about the status? Should I remove that?

Hm, you could move it in a separate Pull Request, then we can merge this one already while we discuss the other topics.

inktrap:
Done!

karrot/groups/models.py
```diff
@@ -63,8 +58,16 @@ def annotate_yesterdays_member_count(self):
         )
 
 
+class GroupManager(BaseManager.from_queryset(GroupQuerySet)):
+    use_in_migrations = True
```
tiltec, line 62: Why is this needed?

PR #972, opened 2020-05-08
Default to enable conflict resolution notifications

nicksellen:
See https://github.com/yunity/karrot-frontend/issues/2043

Open question for now: should I create a db migration to enable that setting for everyone?

I think yes. _But_ it does mean anybody that explicitly set it to off (e.g. they maybe turned it on for a while, then turned it off) will have their preference ignored. _Buuuut_ I suspect that is very few cases, as there are not many issues in the first place.


tiltec:
I think having the migration would be useful - then we can expect everyone to be notified in the rare case of a conflict resolution.

PR #975, opened 2020-05-10
Auto dependency updates

nicksellen:


.circleci/config.yml
```diff
@@ -141,6 +136,43 @@ jobs:
...
+            git checkout -b $BRANCH
+            pip-compile --upgrade
+            git add requirements.txt
+            pushd mjml
+            yarn upgrade --latest
+            git add package.json yarn.lock
+            popd
+            git commit -m "[CI] update dependencies" || exit 0
+            git push --force origin $BRANCH
+            hub pull-request -m "[CI] Update dependencies" -F - <<EOL || true
```
tiltec, line 165: This seems confusing, either use `-m` or `-F`: https://hub.github.com/hub-pull-request.1.html

nicksellen: Yup, good catch! Will update.

PR #1012, opened 2020-06-21
Switch formatter to black

nicksellen:
We currently have yapf, but on our codebase it's exceedingly slow.

Here is is when there are no changes needed:

```
time ./yapf.py 
yapf -i -r -p -e karrot/*/migrations karrot

real	1m8.622s
user	1m40.013s
sys	0m0.382s
```

Whereas black is superfast:

```
time black .
All done! ✨ 🍰 ✨
290 files left unchanged.

real	0m0.271s
user	0m0.209s
sys	0m0.061s
```

... black also feels like what people are using more, the tool feels more polished, nice emoji outputs, friendly options, I think it has good editor support, etc...

Biggest downside is that sometimes the formatting is a bit overly spacious/ugly/etc... and it's not really configurable. For some particularly ugly bits it might be manageable to slightly rewrite the code.

This is just a proposal to see what y'all (I guess primarily @tiltec) would think. I think I would be in favour of moving overall, due to the improved workflow enabled by the speed of processing.


karrot/groups/stats.py
```diff
@@ -7,119 +7,150 @@
...
+                "count_active_editors_{}d".format(
+                    n
+                ): active_memberships.editors().count(),
+                "count_pickup_active_{}d".format(n): pickup_active_memberships.count(),
+                "count_pickup_active_newcomers_{}d".format(
+                    n
+                ): pickup_active_memberships.newcomers().count(),
+                "count_pickup_active_editors_{}d".format(
+                    n
+                ): pickup_active_memberships.editors().count(),
```
nicksellen, line 151: This is one of the ugliest bits.... but changing code to:
```py
fields["count_active_{}d".format(n)] = active_memberships.count()
# ..etc, or f strings:
fields[f"count_active_{n}d"] = active_memberships.count()
```

nicksellen:
I think the pre-commit hook has solved my initial issue (that using yapf was so slow/annoying that I wouldn't do it before commit/push, then it would fail on the server..), so I will close this one. I think it would still be fine to switch, but doesn't seem important now.

PR #1020, opened 2020-07-03
Add activity reminders

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/1027 (sort of)

I first thought to do it like the daily activity notifications, which are in the evening, and send a summary of anything interesting.

But then I realised it wouldn't fit into a push notification message easily. And maybe it's actually more like a reminder.... for an individual activity. So I did it like that!

Right now I implemented it so there are push notifications for all pickups 3 hours before they are due to start.

I used the huey scheduling mechanism, to choose the time in advance, rather than keep polling for which notifications are due. I thought about it for a while, and it wasn't super clear, but I think it should work! We cancel the task if the user steps back from the activity, by keeping the huey task id in the db alongside the activity participant info.

It would be better to make the 3 hour reminder configurable, optional, etc... I think for this the huey scheduling mechanism should work nicely, as we can lookup the options at the time the user joins (.. although would also have to revoke them if the settings are changed...). ... anyway that's for the future.


karrot/activities/receivers.py
```diff
@@ -64,6 +67,29 @@ def remove_activity_participant_from_conversation(sender, instance, **kwargs):
...
 
 
+@receiver(post_save, sender=ActivityParticipant)
+def schedule_activity_reminder(sender, instance, **kwargs):
+    participant = instance
+    if participant.reminder_task_id:
+        return
+
+    activity = participant.activity
+    remind_at = activity.date.start - timedelta(hours=3)
```
nicksellen, line 77: Should check this is actually in the future before scheduling the task.

karrot/activities/receivers.py
```diff
@@ -64,6 +67,29 @@ def remove_activity_participant_from_conversation(sender, instance, **kwargs):
...
+    task = tasks.activity_reminder.schedule(
+        (participant.id, ),
+        eta=remind_at,
+    )
+    participant.reminder_task_id = task.id
+    participant.save()
```
nicksellen, line 83: This might be nice to save without triggering the receiver signal again... I accidently got it into a loop once when I changed the task to clear the `reminder_task_id` after it ran, but then it endlessly rescheduled itself... oops! A bit too easy to trigger that...

karrot/activities/tasks.py
```diff
@@ -1,19 +1,89 @@
...
+        where = ', '.join(part for part in [activity.place.name, activity.place.address] if part)
+
+        title = _('Upcoming %(activity_type)s') % {'activity_type': activity_type}
+        title = f'{emoji} {title}!'
+
+        body = Truncator(' / '.join(part for part in [when, where, activity.description] if part)).chars(num=1000)
+
+        click_action = frontend_urls.activity_detail_url(activity)
+
+        notify_subscribers_by_device(
```
nicksellen, line 76: So far, we do all the push notifications in the subscriptions "app", but I think it's better to have them triggered in the app that triggered them....

nicksellen:
Hmm tests fail in CI, but pass locally, using pytest and normal tests. Hmm.

PR #1022, opened 2020-07-14
Return timezone offsets in activity-related dates

nicksellen:
This implements the API part of https://github.com/yunity/karrot-frontend/issues/959.

This handles it by:
- adding `timezone` annotations to `Activity` and `ActivitySeries` querysets by default
- in the serializers overriding `get_attribute` to return dates with that timezone using `.astimezone(tz)`
- having a custom date time serializer field that actually uses the timezone in the value (the default one gets the current timezone, or a default specified when you create the field)
- fixing the anomaly for the feedback api (as the default timezone annotation for Activity model doesn't apply there...), it's a bit manual, but also keeps the number of queries the same

It doesn't feel very elegant as that is a lot of places to have to consider it, rather than being able to set it in a magic place where everything Just Works.

It might be there is a better way, like setting the timezone for the request context for the current group. Although it would still need to custom field formatters, and it would limit us to one timezone per request, which is probably a reasonable assumption at the moment, but it could make any API requests involve multiple groups complicated at some point.

In the way I did it, theoretically the timezone information could come from elsewhere should we ever want that (e.g. the places could have an override one... not that we have any groups that cross timezones currently).

For anything not activity related, I left it as it is. I think this only makes sense for things that physically happen in the group timezone.

I am assuming the frontend JustWorks at the moment, as date parsing should be fine... the dates still refers to the same point in time.

### DIFF
--- a/karrot/activities/api.py
+++ b/karrot/activities/api.py
@@ -1,3 +1,4 @@
+from django.db.models import F
 from django_filters import rest_framework as filters
 from rest_framework import mixins
 from rest_framework import viewsets
@@ -11,7 +12,7 @@ from karrot.conversations.api import RetrieveConversationMixin
 from karrot.history.models import History, HistoryTypus
 from karrot.activities.filters import (ActivitiesFilter, ActivitySeriesFilter, FeedbackFilter)
 from karrot.activities.models import (
-    Activity as ActivityModel, ActivitySeries as ActivitySeriesModel, Feedback as FeedbackModel
+    Activity as ActivityModel, ActivitySeries as ActivitySeriesModel, Feedback as FeedbackModel, Activity
 )
 from karrot.activities.permissions import (
     IsUpcoming, HasNotJoinedActivity, HasJoinedActivity, IsEmptyActivity, IsNotFull, IsSameParticipant,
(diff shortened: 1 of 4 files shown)


karrot/activities/serializers.py
```diff
@@ -26,15 +28,34 @@ class Meta:
         fields = '__all__'
 
 
+class DateTimeFieldWithTimezone(DateTimeField):
```
nicksellen, line 31: This is maybe better named `DateTimeFieldWithOffset`

nicksellen:
One idea for API updates is to allow it to handle naive datetime values, so that:
```py
value = '2020-07-14 20:30:00'  # look no offset just "20:30" (as if we received this from the client)
naive_date = parse_datetime(value)
tz = pytz.timezone('Europe/Berlin')  # would get the timezone from the relevant group
date = tz.localize(naive_date)  # 20:30 in Berlin, aha, now we know which point in time this is about
```
Otherwise the frontend is going to have some complication to work out the correct point in time to save for an arbitrary date value (without the proper timezone database it cannot know what a date in it's timezone will be in another timezone, the server is cleverer though so could have a utility service to find out...). Although as the linked issue says, we could just forbid editing if you have a different timezone.

karrot/activities/api.py
```diff
@@ -63,10 +64,15 @@ def get_queryset(self):
...
             .select_related('about') \
-            .prefetch_related('about__activityparticipant_set', 'about__feedback_given_by')
+            .prefetch_related('about__activityparticipant_set', 'about__feedback_given_by') \
+            .annotate(timezone=F('about__place__group__timezone'))
         feedback = self.paginate_queryset(queryset)
 
-        activities = set(f.about for f in feedback)
+        activities = set()
+        for f in feedback:
+            activity = f.about
```
tiltec, line 73: I guess this is needed because `selech_related` doesn't use your custom queryset? And it works for activities that come with conversations because that uses `prefetch_related`?

nicksellen: > I guess this is needed because selech_related doesn't use your custom queryset?

Indeed. Which seems a shame, but also how it is.

> And it works for activities that come with conversations because that uses prefetch_related?

Not sure how it relates/impacts conversations, I didn't change anything there. But I guess yeah that prefetch_related bit ensures it has enough information for the serializer?

PR #1024, opened 2020-07-17
Allow offers without picture

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/2067 (backend part)

(also sneaks in defaulting groups to having offers features enabled)


karrot/groups/models.py
```diff
@@ -91,7 +95,7 @@ class Group(BaseModel, LocationModel, ConversationMixin, DirtyFieldsMixin):
         upload_to='group_photos',
         null=True,
     )
-    features = ArrayField(TextField(), default=list)
+    features = ArrayField(TextField(), default=default_group_features)
```
tiltec, line 98: Just wondering, does this function really get called on group creation?

karrot/offers/serializers.py
```diff
@@ -86,20 +86,6 @@ def update(self, instance, validated_data):
...
-        add_image_count = 0
-        remove_image_count = 0
-        for image in images:
-            if image.get('_removed', False):
-                remove_image_count += 1
-            elif not image.get('id', False):
-                add_image_count += 1
-        resulting_image_count = existing_image_count + add_image_count - remove_image_count
-        if resulting_image_count == 0:
-            raise serializers.ValidationError(_('Must have at least one image'))
```
tiltec, line 100: Might need to run `makemessages` to update the PO files!

karrot/groups/models.py
nicksellen: Yes, well I run it locally and created a group and it had offers enabled. I had noticed it because I needed a group with offers by my new local groups didn't have offers enabled.

PR #1028, opened 2020-07-20
Add images to conversation messages

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/1793#issuecomment-661051412 (backend version)

This adds image uploads to conversation messages. It uses the same implementation as offer images.

- messages can have 0 or more images
- they are uploaded using the multipart JSON parser thing I implemented
- it keeps "full size" and 200x200 thumbnail (the frontend limits the full size size...)
- I added all the prefetches so the query count should just be 1 higher than before, not 1-per-message higher


nicksellen:
> I recommend to avoid using full_size images in frontend, they can behave different from thumbnails as they are in fact the original image (as delivered by the API). For example, if we decide to use a different size in Croppa, full_size will become inconsistent. Better use thumbnails in 200x200 for preview and 600x600 for detail view. full_size might be good for mega-high-res if wanted...

Yup, actually I realised I was not so happy with forcing the images to be square, but vue-croppa doesn't really support natural aspect ratio (I guess that's the point of the component...), so my thinking was keeping the full size means we don't need to encode the aspect ratio of the enlarged image in the backend, but can evolve the frontend as needed (whereas for thumbnails I think forcing a square is reasonable) - in the frontend I make the images fit into where I want them to go regardless of their actually size, so it should stay consistent with historic sizes.

Although, I _think_ if using the "thumbnail" sizer with versatile image field it will not crop it to a square, but merely ensures it would fit inside a box of that size (so do a "longest edge" kind of thing if specified as a square), from [the docs](https://django-versatileimagefield.readthedocs.io/en/latest/using_sizers_and_filters.html): 
- thumbnail sizer: _Here’s how you would create a thumbnail image that would be constrained to fit within a 400px by 400px area:_
- crop sizer: _To create images cropped to a specific size, use the crop Sizer:_

... and that would be a desirable behaviour, to preserve the aspect ratio of the uploaded image, should we change it later.

I'll add a 600x600 thumbnail sizer and that should be enough for now :)

tiltec:
> but vue-croppa doesn't really support natural aspect ratio (I guess that's the point of the component...)

Indeed, I forgot that limitation - might be good if users could upload uncropped pictures. Maybe we need another image upload component for this.

nicksellen:
> Indeed, I forgot that limitation - might be good if users could upload uncropped pictures. Maybe we need another image upload component for this.

Yup, I had a little look for one, but nothing was obvious, and the backend should be able to stay the same in any case :)

nicksellen:
Ok I added:
- message editing tests with images
- 600x600 thumbnail for message images

PR #1040, opened 2020-08-14
Add activity history statistics endpoint

nicksellen:
Discussion https://community.foodsaving.world/t/more-statistics-display/489
Frontend PR https://github.com/yunity/karrot-frontend/pull/2163

Activity history endpoint!

- because of the "feedback as sum" vs "feedback as average" complication, I removed unified that, so it's always as sum now (only historic data had it as average >6 months old), the old value is still stored in another (unused) column, but we could remove that too at some point
- queries on the history table
- it reports on left activities, but _only_ if that activity was not later done and full (the idea here is to not report possibly negative information about people if the activity got done anyway... maybe they just managed to find a last minute replacement, who knows...)
- has user, place, group, and date range filter options

~~I thought I'd just start with an API endpoint and see how it feels. It has a pretty gnarly query to get the count of late dropouts, but it seems to work!~~

~~Even though the stats are about activities, it's actually based on the place model, so I named the endpoint that. I would imagine different endpoints for different statistics reports... perhaps this one should be qualified further in the url, as there could be other place reports.~~ I switched it to query based on history model in the end, made the query easier.

I'm thinking one viewset/serializer combo per "report", so we get all the parameters properly in swagger, etc..


tiltec:
Nice, especially the query :+1: 

There's some overlap with `/api/places/:id/statistics/`, both are stats about one place: https://github.com/yunity/karrot-backend/blob/dc639d58df533f72d4054643dbebc322f11ce663/karrot/places/api.py#L66-L81

This endpoint gets called when visiting a places page, so it shouldn't be too heavy on the database. Maybe we want to have two levels of place reports, "overview" and "drill-down"?

nicksellen:
> There's some overlap with /api/places/:id/statistics/, both are stats about one place

The one I just added here gives you a _list_ not a single place, like from here:

 

(except with all the relevent activity history types returned, so can have multiple columns rather than have to select type)

> This endpoint gets called when visiting a places page, so it shouldn't be too heavy on the database. Maybe we want to have two levels of place reports, "overview" and "drill-down"?

I was thinking to have them inside the stats "app" as they will not all be about places, and might actually be about other things that don't directly relate to a model. For these ones that I just created the user will actually be seeing it as an activity report for a _user_ rather than a _place_ report (as I was imaging the UI flow to be like **statistics/reports -> activities -> user** kind of thinking), it kind of just happens that breaking it down by place makes sense here.

tiltec:
Ah yeah, that makes sense!

tiltec (approved):

PR #1042, opened 2020-08-15
Fixup transaction handling for message images

nicksellen:
For the offer images I had to ensure to do some of notifications using `transaction.on_commit` to ensure the images were saved and available for the receiver handlers/notifications/etc, we also need to do this for message images so. I implemented an `@thing` that simplifies it now for both cases, and made the tests TransactionTestCases where applicable.

~~... actually I might tag a few other things into this PR that relate to the message images ...~~ changed my mind

### DIFF
--- a/karrot/offers/receivers.py
+++ b/karrot/offers/receivers.py
@@ -1,19 +1,17 @@
-from django.db import transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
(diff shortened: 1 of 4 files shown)


nicksellen:
> Looks good! I guess the tests didn't fail because they don't check for images in messages?

Indeed, that's a bit crappy of me not to add the tests for that... but ...

PR #1044, opened 2020-08-18
Order feedback by activity start date

nicksellen:
Fixes https://github.com/yunity/karrot-frontend/issues/2157

The cursor pagination disallows double-underscore lookups for pagination ordering, I think because it assumes they will break the "rules" on what can be a valid ordering field, see [docs](https://www.django-rest-framework.org/api-guide/pagination/#details-and-limitations). In our case I think this should be fine still, as by the time the feedback is created the pickup cannot be changed anyway as it's in the past. I cheated the check by using an annotation...


tiltec:
Seems fine, I think I used the annotation trick somewhere else too!

PR #1045, opened 2020-08-18
Custom activity types

nicksellen:
Based on discussion at https://community.foodsaving.world/t/custom-activity-and-place-types/496.

It is still more like a concept for now than a solution. Could still change a lot. A few points:
- activity types are _per group_ (so each group would have it's _own_ "pickup" type)
- the migration creates a "task" type for each group, and a "pickup" type for foodsaving groups
- all existing pickups are set to "pickup" (for foodsaving groups) or "task" (for any other group types)
- there is one API endpoint so far `/api/activity-types/` that returns a list of the activity types and all the fields
- I would leave out anything fancy like custom fields in this PR
- _possibly_ it could make sense to get this PR merged (assuming we like the concept), without making it possible to add/edit/delete types, to make it more incremental... but let's see

One of the bits I didn't think so much about right now is the statistics. If each group can have totally custom types, we have no magic way of seeing how many "pickups" are done across the whole software, unless they leave the name unchanged.

I did consider having some "global" activity types, so we would still know some stats between groups, but I prefer the idea each group can truly customize itself.

(as a sidenote about statistics, I wondered if at some point we should just store them directly in the postgres db instead of influxdb, as grafana can already query postgres, and if the performance was a big problem https://www.timescale.com/ is a solution to that - an extension that runs inside your existing postgres installation and makes the tables magic and fast, but probably for our case we wouldn't even need that... /sidenote)


nicksellen:
Closing in favour of https://github.com/yunity/karrot-backend/pull/1063

PR #1046, opened 2020-08-28
Remove some playground exceptions

tiltec:
Related to https://github.com/yunity/karrot-frontend/pull/2175 and https://community.foodsaving.world/t/how-to-handle-the-spam-in-the-playground-group/486

- Don't make new playground group members editors by default
- Don't prevent bell notifications when playground group member became editor


nicksellen (approved):
Looks good!

PR #1063, opened 2020-10-12
Add custom activity types

nicksellen:
- discussion: https://community.foodsaving.world/t/custom-activity-and-place-types/496
- frontend PR: https://github.com/yunity/karrot-frontend/pull/2207

It is still more like a concept for now than a solution. Could still change a lot. A few points:
- activity types are _per group_ (so each group would have it's _own_ "pickup" type)
- the migration creates a "task" type for each group, and a "pickup" type for foodsaving groups
- all existing pickups are set to "pickup" (for foodsaving groups) or "task" (for any other group types)
- there is one API endpoint so far `/api/activity-types/` that returns a list of the activity types and all the fields
- I would leave out anything fancy like custom fields in this PR
- _possibly_ it could make sense to get this PR merged (assuming we like the concept), without making it possible to add/edit/delete types, to make it more incremental... but let's see

One of the bits I didn't think so much about right now is the statistics. If each group can have totally custom types, we have no magic way of seeing how many "pickups" are done across the whole software, unless they leave the name unchanged.

I did consider having some "global" activity types, so we would still know some stats between groups, but I prefer the idea each group can truly customize itself.

(as a sidenote about statistics, I wondered if at some point we should just store them directly in the postgres db instead of influxdb, as grafana can already query postgres, and if the performance was a big problem https://www.timescale.com/ is a solution to that - an extension that runs inside your existing postgres installation and makes the tables magic and fast, but probably for our case we wouldn't even need that... /sidenote)


karrot/activities/models.py
```diff
@@ -18,6 +19,24 @@
 from karrot.places.models import PlaceStatus
 
 
+class ActivityType(BaseModel):
+    group = models.ForeignKey('groups.Group', on_delete=models.CASCADE, related_name='activity_types')
+    name = models.CharField(max_length=80)
+    name_is_default = models.BooleanField(default=True)
```
nicksellen, line 25: Maybe `name_is_translatable` / `name_is_translated`?

karrot/activities/serializers.py
```diff
@@ -140,6 +147,20 @@ def validate_date(self, date):
...
+        def get_instance_attr(field):
+            if self.instance is None:
+                return None
+            return getattr(self.instance, field)
+
+        typus = data.get('typus', get_instance_attr('typus'))
+        place = data.get('place', get_instance_attr('place'))
+
+        if typus and place and typus.group_id != place.group_id:
+            raise serializers.ValidationError(_('Typus is not for this group.'))
```
nicksellen, line 160: Don't translate strings that are not helpful for the user.

karrot/activities/tests/test_tasks.py
```diff
@@ -54,7 +54,7 @@ def test_activity_reminder_notifies_subscribers(self, notify_subscribers_by_devi
             kwargs['click_action'],
         )
         self.assertIn(
-            'Upcoming pickup',
+            'Upcoming {}'.format(self.activity.typus.name),
```
nicksellen, line 57: Some languages might have different noun forms so this kind of translation is too limited.

karrot/groups/serializers.py
```diff
@@ -6,12 +6,14 @@
...
 from versatileimagefield.serializers import VersatileImageFieldSerializer
 
+from karrot.activities.activity_types import default_activity_types
 from karrot.groups.models import Group as GroupModel, GroupMembership, Agreement, UserAgreement, \
     GroupNotificationType
 from karrot.history.models import History, HistoryTypus
 from karrot.utils.misc import find_changed
 from karrot.utils.validators import prevent_reserved_names
 from . import roles
+from ..activities.models import ActivityType
```
nicksellen, line 16: Change import style

karrot/activities/activity_types.py
```diff
@@ -0,0 +1,28 @@
+from django.utils.translation import gettext as _
```
nicksellen, line 1: Maybe I need to use lazy version?

karrot/activities/activity_types.py
```diff
@@ -0,0 +1,28 @@
...
+
+default_activity_types = {
+    'Pickup': {
+        'name': _('Pickup'),
```
nicksellen, line 5: Could/should/maybe store these names outside/seperate from the different types, as they are slightly independent from the defaults...

karrot/management/commands/foo.py
```diff
@@ -0,0 +1,27 @@
+from django.core.management import BaseCommand
```
nicksellen, line 1: No foo here please!

karrot/activities/models.py
```diff
@@ -43,6 +62,12 @@ class ActivitySeries(BaseModel):
     description = models.TextField(blank=True)
     duration = DurationField(null=True)
 
+    typus = models.ForeignKey(
```
nicksellen, line 65: Name is `activity_type`? Is it consistent?

PR #1065, opened 2020-10-19
Add geoip for group preview endpoint

nicksellen:
Users get a bit confused sometimes when signing up and lose their way to the group they were trying to get to. We can make this a bit easier by showing them the groups closest to them more prominently.

It's mainly a frontend task, but in order to support that we can use geoip to calculate the distance of the groups to the user. This adds an extra "distance" field to the group preview data with the distance information in km, rounded to the nearest km.

It will work whether the user is logged in or not. If they are logged in, the lat/lng result will be saved in the session for future use.

Now I'm writing this I'm wondering if it's also useful to return the geoip value to the user too, so it can be used for centring the map, etc... (could even have calculated the distances fully in the client if so...).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
     docker:
-      - image: karrot/python:2.3
+      - image: karrot/python:2.4
         environment:
           PGHOST: 127.0.0.1
       - image: circleci/postgres:11
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: karrot/python:2.3
+    - image: karrot/python:2.4
 
 aliases:
   - &restore-env-cache
--- a/karrot/groups/serializers.py
+++ b/karrot/groups/serializers.py
@@ -280,7 +280,7 @@ def get_client_ip(request):
         return request.META.get('REMOTE_ADDR')
 
 
-@lru_cache
+@lru_cache()
 def ip_to_lat_lon(ip):
     try:
         return geoip.lat_lon(ip)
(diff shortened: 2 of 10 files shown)


tiltec (commented):
Nice!

Yeah, at some point the frontend might be interested in the IP geolocation coordinates as well - then I guess we could create another endpoint for this.

I wonder if IPs can change while the session stays the same?
In any case, it might be simpler to use a [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) instead of the session.

How should we deploy and update the GeoIP database? Karrot itself could take care of it with a  huey task. or it could be outside, getting it in ansible and maybe updating via system cronjob.

nicksellen:
Yup, lru_cache is a good idea. I think it's probably only edge cases where it matters a lot, as it's only an approx location anyway, but yeah.

I didn't think much about the db, but with ansible makes sense to me, not sure how automatable it is though as we have to be signed in, didn't investigate that bit much yet.

I do wonder whether to make the API already for the frontend to receive the geoip co-ordinates... although the way I implemented the frontend so far it doesn't need it (focuses the map on the closest few places).

nicksellen:
Ok, that uses lru_cache now instead, was pretty simple quick change, and the code is simpler too. I left it at the default maxsize of 128.

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: karrot/python:2.3
+    - image: karrot/python:2.4
```
nicksellen, line 4: Check karrot docker to make sure it has the geoip libs installed.

karrot/groups/serializers.py
```diff
@@ -257,13 +263,94 @@ def update(self, instance, validated_data):
...
+    Returns distance of the object from users current location.
+    - the object must have latitude/longitude fields
+    - the users location is detirmined via geoip if available
+    - the geoip lookup is cached in an lru_cache
+    - the return unit is km rounded to the nearest km
+
+    It may return None under various conditions:
+    - the GeoIP2 libary was not initialized (missing the files)
+    - there is no request context
+    - we cannot detirmine the IP address of the client
```
nicksellen, line 303: detirmine -> determine

PR #1067, opened 2020-10-23
activity: avoid crossing limit when adding participants concurrently

tiltec:
A race condition can happen when two people try to join an activity at the same time. https://chat.foodsaving.world/channel/karrot-dev?msg=BpBqKd9nC5M5aWMKG was a recent report about this problem.

This PR fixes it by locking the activity while we attempt to add a participant. It seems hard to test this reliably, so in the end I omitted the test.

See also https://github.com/yunity/karrot-backend/issues/409 for the more general problem.

### DIFF
--- a/karrot/activities/api.py
+++ b/karrot/activities/api.py
@@ -1,3 +1,4 @@
+from django.db import transaction
 from django.db.models import F
 from django_filters import rest_framework as filters
 from rest_framework import mixins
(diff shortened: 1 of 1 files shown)


nicksellen (approved):
Looks nice and simple! And yes, hard to test.. let's go for it.

PR #1072, opened 2020-10-31
Add bootstrap endpoint

nicksellen:
Adding a `/api/bootstrap/` to return some useful initial data for the frontend, geoip, auth user, and public groups info. Used by https://github.com/yunity/karrot-frontend/pull/2212.

It also randomly refactors status endpoint to use StatusSerializer, as I did play with a version that included status data, but reverted that bit, but thought might as well keep the StatusSerializer...

Also randomly adds a `member_count` field to the group previous, so we can eventually remove the member list...

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -87,6 +87,7 @@ INSTALLED_APPS = (
     'karrot',
     'karrot.applications.ApplicationsConfig',
     'karrot.base.BaseConfig',
+    'karrot.bootstrap.BootstrapConfig',
     'karrot.community_feed.CommunityFeedConfig',
     'karrot.issues.IssuesConfig',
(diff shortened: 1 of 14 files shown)


tiltec (approved):
Seems good! 

Possibly add a boolean `is_member` field to the group info serializer, as this is needed to eventually remove the `members` list. But this isn't the goal of this PR anyway, so we could also delay it.

nicksellen:
Ok, I added an `is_member` field! Although then I wondered whether that is even enough for removing the member list, we could supply the member list if the user is a member...

By adding `is_member` I might have added an n+1 query though... well also the members count field might not be so great to run for each entry... hmmm.

PR #1081, opened 2020-11-18
Add activity type editing API

nicksellen:
Frontend PR: https://github.com/yunity/karrot-frontend/pull/2240

I think this is enough for a first round now.

It adds editing capabilities for activity types and websocket updates.

It also adds a `status` field for activity types, which can be `active` or `archived`. If it's `archived` then creation of new activities or series will be forbidden, and the series won't be included when generating new activities (existing ones will be left as they are).

It does not implement what we discussed about propagating the `has_feedback` field to the individual activities. This would be the way to support more sophisticated mechanisms when making modifications to the `has_feedback` option (e.g. "only change future activities"), which I think might be useful, but I would prefer to go forward with the simplest version for now, and wait until there is a clearer need for that behaviour.

I also used the opportunity to evolve the history model by adding a "message" field, intended to contain something that a human writes to explain a change. In my frontend work, I have implemented this as a dialog that opens when you press save, and it asks you to optionally write a message describing or explaining the change, and displays it in the history view.

I had wanted to also be able to show a history of the specific activity type, but I think the history mechanism needs a bit deeper reworking to support that properly, so I would leave that for now.

I also added the `before` and `after` fields into the history API serializer, which was useful for displaying the "activity type change" history entry... as I didn't add another foreign key on the history model. I would like to clarify a bit more how we intend to use it... and maybe this bigger reworking... but it's more something to make it useful enough for now, without too much change.

## TODO

- [x] websocket updates for changes
- [ ] ~~maybe a bell notification for new types?~~
- [x] history entries for create/edit/delete
- [x] (more)tests
- [x] more permissions?
- [x] group only settable during creation
- [x] only can choose your own groups (via group editor permission thing)


karrot/history/serializers.py
```diff
@@ -18,6 +18,9 @@ class Meta:
             'place',
             'users',
             'payload',
+            'message',
+            'before',
```
tiltec, line 22: Did you check if the values in `before` and `after` are always good to return? They contain all fields of the related model,

tiltec (approved):
Code looks good! From my perspective, it's good to merge. Especially now that you released activity types, we have some more testing time in dev.

karrot/history/serializers.py
nicksellen: Ah good point. I did not check. Maybe we can run it through the serialiser. Hmm I'll have a think.

Another way would be only returning it for my new types, adding a bit of tech debt until we can rationalize the history data.

nicksellen: It already used the serializer to write into the `after` field (I presume others do that too), but I decided to make it so only the new specific history types will return the `after` field (and the `before` field was actually unused in my frontend code, so I removed that entirely).

PR #1104, opened 2021-01-05
Improve notifications when somebody left the group (#1619)

mvellasco:
## Issue
https://github.com/yunity/karrot-frontend/issues/1619

## Description
* Added new user queryset filter condition to match users that have GROUP_LEAVE histories in one of current user's groups


nicksellen:
(I copied in the comment from https://github.com/yunity/karrot-frontend/issues/1619#issuecomment-754326542 to keep the discussion to the PR, seems best!)

> @nicksellen Right now I'm not in any groups, I'm trying to set up one though in my local town(Campos dos Goytacazes - Rio de Janeiro - Brazil). About the solution, you guys had in mind to copy the user data into the history model, it might be even simpler if I didn't overlook anything. How about instead of adding a new DB field for that we just add another filter to the query for users matching if they have a GROUP_LEFT type history in the current user's group? I took the liberty to make a draft implementation and no tests broke so it should be fairly safe. In any case, let me know so I can submit the PR alongside the bell notification as well. Thanks!

Thanks! I think it's a good idea. I think it might be worth adding a grace period though, maybe only history entries within 3 or 6 months (@dwaxweiler [mentioned](https://github.com/yunity/karrot-frontend/issues/1619#issuecomment-628089182) 2 weeks but seems a bit short to me)? If you leave a group I think it's reasonable that your information is no longer visible at some point.

One downside of doing it this way is that if we wanted to just restrict it to their name (but hide the email address) we don't have a way of limiting visibility for the left users.

But this PR improves the current behaviour so seems good to me.

Could you implement a test?

karrot/users/api.py
```diff
@@ -52,8 +53,12 @@ def get_queryset(self):
 
         groups = self.request.user.groups.all()
         is_applicant_of_group = Q(application__group__in=groups)
+        has_left_group = Q(
+            history__group__in=groups,
+            history__typus=HistoryTypus.GROUP_LEAVE,
```
nicksellen, line 58: I always struggle to remember exactly what happens when putting multiple deeper (ones with a `__`) conditions in a `Q` query. I think there are some caveats where sometimes it doesn't quite do the intended query and the conditions are not applied to different row or something.

Sorry, I'm a vague on the exact details. I usually end up checking the SQL when I'm not sure (using `print(some_queryset.query)` in a Django shell and pretty printing it, I can understand SQL better than Django ORM :) ) ... also tests are a nice thing too for that.

mvellasco:
> (I copied in the comment from [yunity/karrot-frontend#1619 (comment)](https://github.com/yunity/karrot-frontend/issues/1619#issuecomment-754326542) to keep the discussion to the PR, seems best!)
> 
> > @nicksellen Right now I'm not in any groups, I'm trying to set up one though in my local town(Campos dos Goytacazes - Rio de Janeiro - Brazil). About the solution, you guys had in mind to copy the user data into the history model, it might be even simpler if I didn't overlook anything. How about instead of adding a new DB field for that we just add another filter to the query for users matching if they have a GROUP_LEFT type history in the current user's group? I took the liberty to make a draft implementation and no tests broke so it should be fairly safe. In any case, let me know so I can submit the PR alongside the bell notification as well. Thanks!
> 
> Thanks! I think it's a good idea. I think it might be worth adding a grace period though, maybe only history entries within 3 or 6 months (@dwaxweiler [mentioned](https://github.com/yunity/karrot-frontend/issues/1619#issuecomment-628089182) 2 weeks but seems a bit short to me)? If you leave a group I think it's reasonable that your information is no longer visible at some point.
> 
> One downside of doing it this way is that if we wanted to just restrict it to their name (but hide the email address) we don't have a way of limiting visibility for the left users.
> 
> But this PR improves the current behaviour so seems good to me.
> 
> Could you implement a test?

I'm glad it'll help! I'll implement the tests and add the grance period logic with the bell ring notification code as well. About that Q object issue, you're correct, it shouldn't have reference to more than one field per Q object, so I'll change that to : 
```python
has_left_group = (
    Q(history__group__in=groups) & Q(history__typus=HistoryTypus.GROUP_LEAVE)
)
```

mvellasco:
@nicksellen About that caveat of showing the email, I can add a conditional on the user viewset like that to mitigate the problem. Please let me know if you think that's a good idea. Thanks! 
```python
def profile(self, request, pk=None):
    user = self.get_object()
    if not any(group in user.groups.all() for group in request.user.groups.all()):
        serializer = UserInfoSerializer(user)
    else:
        serializer = self.get_serializer(user)
```

nicksellen:
Great!

I like the idea of using the serializer to show different info as you did, a complexity is that because we keep the frontend up to date (via signals and websockets) we have to send the updated information when memberships change. it _might_ JustWork :tm: but also might not... there are some issues there already.

I think it also might cause a lot of queries, as it's per user, although ensuring that the groups are prefetched in the original query should mitigate that.

On balance I think maybe leaving that out of this PR is fine, the need for it might just be in my head after all!

I'll await the other changes you mentioned, and can do a more in depth code review :)

... oh, and regarding the bell notification for user leaving, does one already exist? (I can't remember, and too lazy to check right now...)

mvellasco:
@nicksellen I've addressed the new bell notification creating when the user has left and the changes on the Q objects query. Also I've created a PR on the frontend as well to properly display the notification info, I'll ping you for a initial review there. Thanks for the patience and sorry it took so long to come back to this, it has been quite a crazy year so far :slightly_smiling_face:

mvellasco:
@nicksellen I've addressed the new bell notification creating when the user has left and the changes on the Q objects query. Also I've created a PR on the frontend as well to properly display the notification info, I'll ping you for a initial review there. Thanks for the patience and sorry it took so long to come back to this, it has been quite a crazy year so far :slightly_smiling_face:

karrot/notifications/receivers.py
```diff
@@ -188,9 +188,20 @@ def new_member(sender, instance, created, **kwargs):
...
     membership = instance
 
+    for user in membership.group.members.all():
+        Notification.objects.create(
+            user=user,
+            type=NotificationType.MEMBER_LEFT.value,
+            context={
+                'group': membership.group_id,
+                'user': membership.user_id,
+                'added_by': membership.added_by_id,
```
nicksellen, line 201: I think we don't need `added_by` here, as this is just about leaving.

nicksellen (changes requested):
Hey! Sorry for the delay in reviewing.

Looks good to me, I'm still a bit uncertain about whether we should already add a time-limit to the visibility, but happy to proceed without.

Just one minor change request about the `context` data for the notification.

Thanks! :star2:

karrot/notifications/receivers.py
mvellasco: hey @nicksellen I've addressed this change, and also added the code for the time limit functionality(60 days). Let me know if that looks good :slightly_smiling_face:

PR #1105, opened 2021-01-16
Add activities with role requirement

nicksellen:
Forum thread: https://community.foodsaving.world/t/applicant-trial-pickup-proposal/575
Frontend PR: https://github.com/yunity/karrot-frontend/pull/2421

Initial work on trial activities. It's a work in progress for now still with open questions and maybe some of this stuff will change.

The concept is:
- activities can have a _role_ requirement
- if they have that, they can also specify >0 max participants who can join without the role (_open participants_)
- participants join with or without the role (assuming there are enough appropriate spaces left)
- there is currently no additional roles or mechanisms to get them, as we are still working that bit out :) (working on that here https://github.com/yunity/karrot-backend/pull/1167)

## TODO

- [x] all the comments I left
- [x] make the activity notification emails take into account the roles
- [ ] resolve the situations where `activity.participants` might need to take into account of role stuff
- [x] also add to activity series
- [x] consider when updating series (require_role, max_open_participants) if we want to update the individual activities... (what do we generally do in that case?)
- [ ] some more stuff probably


karrot/activities/tests/test_activities_api.py
```diff
@@ -497,6 +500,95 @@ def test_export_ics_activities_as_group_member(self):
...
+        self.assertEqual(participant.role, APPROVED)
+
+    def test_backwards_compatible_participants_api(self):
+        self.activity.add_participant(self.member, role=None)
+        self.activity.add_participant(self.approved_member, role=APPROVED)
+        self.client.force_login(user=self.member)
+        response = self.client.get('/api/activities/{}/'.format(self.activity.id))
+
+        # participants field only shows participants without roles user ids
+        self.assertEqual(response.data['participants'], [self.approved_member.id])
```
nicksellen, line 557: This test no longer does what it says it does, as the existing frontend won't send the `role` parameter.

karrot/activities/tests/test_activities_api.py
```diff
@@ -497,6 +500,95 @@ def test_export_ics_activities_as_group_member(self):
...
+            user=self.approved_member,
+            roles=[APPROVED],
+        )
+        self.activity = ActivityFactory(
+            place=self.place,
+            require_role=APPROVED,
+            max_open_participants=1,
+        )
+
+    def test_cannot_join_if_requires_role_and_none_without_role(self):
```
nicksellen, line 520: Can rename this to refer to "open" participants

karrot/activities/tests/test_activities_api.py
```diff
@@ -9,11 +10,13 @@
 from karrot.groups.factories import GroupFactory
 from karrot.groups.models import GroupMembership, GroupStatus
 from karrot.activities.factories import ActivityFactory, ActivityTypeFactory
-from karrot.activities.models import to_range
+from karrot.activities.models import to_range, ActivityParticipant
 from karrot.places.factories import PlaceFactory
 from karrot.tests.utils import ExtractPaginationMixin
 from karrot.users.factories import UserFactory
 
+APPROVED = 'approved'
```
nicksellen, line 18: How/where should the roles be defined?

karrot/activities/serializers.py
```diff
@@ -391,25 +416,31 @@ def get_attendee(self, activity):
...
+        ]
+
+    role = CharField(write_only=True)
 
     def update(self, activity, validated_data):
         user = self.context['request'].user
-        activity.add_participant(user)
+        place = activity.place
+        group = place.group
+        activity.add_participant(user, validated_data.get('role', None))
```
nicksellen, line 429: I think need a new permission check to make sure the user actually has that role, either here, or in a permission class.

karrot/activities/templates/activity_notification.mjml
```diff
@@ -19,12 +19,14 @@
...
                                         </a>
+                                        {# TODO maybe show if it requires a role or not... #}
                                         {% if show_empty_participants_count %}
                                             -
-                                            {% trans trimmed empty_participants_count=activity.empty_participants_count(),
+                                            {% trans trimmed empty_participants_count=activity.empty_participants_count(role=activity.require_role),
                                                     max_participants=activity.max_participants %}
                                                 {{ empty_participants_count }}/{{ max_participants }} remaining
                                             {% endtrans %}
+                                            {# TODO maybe show the available slots without role remaining too #}
```
nicksellen, line 29: These TODOs would be useful to do!

nicksellen: Well, this is a little bigger now, as we should take into account the roles of the user in which ones we show...

karrot/activities/tests/test_activities_api.py
nicksellen: Can be out of scope for this PR...

nicksellen:
There are quite a few areas where `activity.participants` is used, and this won't take into account any of the role stuff. I've used the term "open participants" in places, to indicate the "trial" type purpose, with a more general term.

But it's not obvious in each usage whether `activity.participants` should include the open participants. It will by default. But that might be wrong. As they are kind of like "bonus" people for an activity.

Needs some more thinking.

nicksellen:
This is getting close! Two things I changed just now:
- instead of tracking the `role` in the `ActivityParticipant`, it just tracks `is_open` - this simplifies a lot of checks
- for the activity notification email, only show activities that match the role requirements of the user (won't show "open" activities)

There is still a bit of word confusion I think though as "activity participants" is sometimes including open participants, sometimes not. I don't have a word for "non-open" participants either.

Maybe also we can have a text field for "open participant description" or something, which would be text that people see when they don't meet the role requirement - which could be used for "trial pickup" instructions.

nicksellen:
Closed in favour of https://github.com/yunity/karrot-backend/pull/1199 (which builds on this)

PR #1114, opened 2021-02-14
Batch influxdb writes

nicksellen:
During a request, if something changes we notify websocket subscribers about we also send out also has a stats write. This is a lot happening. As the number of websocket updates scales with connected users/sessions.

I am wondering if this is a reason sometimes requests are very slow.

This is not the solution to everything (I suspect we should push those websocket update tasks to a backend huey task), but still, this might improve things! In my local dev environment it was able to batch ~20 writes that would have been 20 seperate http requests made during in a single web request to karrot (!).

(The client for influxdb 2 has built in batching, but influxdb 2 upgrade needs a requires migration so I didn't want to go into that... and I am wondering about storing stats in postgres... but that's a much bigger task).


tiltec (approved):
Looks good to me! I guess we expect POST requests to be faster, but not much change in GET requests.

PR #1119, opened 2021-03-02
Add uploads folder

pogopaule:
An uploads folder is required. If it is not present an error occurs,
which is not so obvious when first setting up the backend.

This solution keeps the uploads folder and ignores its content.

See: https://stackoverflow.com/a/5581995/776800


nicksellen (approved):
Thanks! I would have tried to include it all in the root .gitignore, not sure if that is possible, but I'll trust this one and approve/merge :)

pogopaule:
The alternative approach is to add a file (e.g. `.gitkeep`) to the `uploads` folder and ignore the contents of it (except for `.gitkeep`)

See https://stackoverflow.com/a/7229996/776800

nicksellen:
Yup, I used that approach before, but this one also seems good!

PR #1122, opened 2021-03-18
Add place types and statuses

nicksellen:
**Update** I'm thinking the work I've done is a bit complicated, so having some thoughts over at https://community.foodsaving.world/t/custom-place-types/705

---

First efforts at the backend for adding place types.

One of the open questions for a while had been how to handle the statuses, the current ones are aimed at store co-operations for foodsaving (negotiating, co-operating, etc...), so I thought maybe having statuses _per place type_ made sense, but then realised some of the statuses would be relevant for different types (maybe types would not just be store, but _bakery_ or _supermarket_), so I went another way, custom statuses too, that can be used for any place, independent of place type.

Sometimes we have different behaviour in the software depending on the status, and that is still useful, so I added the concept of "category" of the status which can be inactive/active/archived, so the person defining a status can indicate the _broad meaning_ of the status, and the software can behave accordingly.

## TODO

- [ ] check if I need to do some logic if places become active because place status category was changed
- [ ] consider simplifying `PlaceType`/`PlaceStatus`/`PlaceStatusCategory`... it might be we can just have active/archived directly on each `Place` and have activities enabled for all non-archived ones... maybe can remove `has_activities` field...


nicksellen:
Closing in favour of https://github.com/yunity/karrot-backend/pull/1159 (but the statuses stuff in this branch might be useful...).

PR #1124, opened 2021-03-23
feat: allow to dismiss feedback

pogopaule:
## Links to related issues

Related frontend PR https://github.com/yunity/karrot-frontend/pull/2319


karrot/activities/permissions.py
```diff
@@ -16,6 +16,13 @@ def has_object_permission(self, request, view, obj):
             return obj.is_upcoming()
 
 
+class IsNotUpcoming(permissions.BasePermission):
+    message = 'The activity is not in the past.'
+
+    def has_object_permission(self, request, view, obj):
```
layla19, line 22: Should we also allow all SAFE_METHODS here to be consistent with IsUpcoming?

larzon83: Please elaborate 😄 🛩️

karrot/activities/tests/test_activities_api_filter.py
```diff
@@ -131,8 +131,16 @@ def setUp(self):
                 self.member,
             ], date=self.oneWeekAgo
         )
+
         self.feedback = FeedbackFactory(about=self.activityFeedbackAlreadyGiven, given_by=self.member)
 
+        self.activityFeedbackDismissed = ActivityFactory(
```
nicksellen, line 137: The consistent (locally), yet inconsistent (globally), casing really bothers me here! Just wanted to say so, not for you to fix, but you can if you want! (even though it's a bit unrelated...).

karrot/subscriptions/receivers.py
```diff
@@ -642,3 +642,19 @@ def feedback_saved(sender, instance, created, **kwargs):
 
     for subscription in ChannelSubscription.objects.recent().filter(user=user):
         send_in_channel(subscription.reply_channel, topic='status', payload=payload)
+
+
+@receiver(post_save, sender=ActivityParticipant)
+def activity_participant_saved(sender, instance, **kwargs):
```
nicksellen, line 648: It would seem worth the small abstraction to me to implement a function: `send_feedback_possible_count(user)` to save the code duplication.

That can be called from the existing handlers `feedback_saved` and `send_activity_participant_updates` (which also handles the case of leaving the activity).

karrot/activities/serializers.py
```diff
@@ -181,13 +182,17 @@ class Meta:
...
+    feedback_dismissed_by = serializers.SerializerMethodField()
     feedback_due = DateTimeFieldWithTimezone(read_only=True, allow_null=True)
 
     date = DateTimeRangeField()
 
     def get_participants(self, activity):
         return [c.user_id for c in activity.activityparticipant_set.all()]
 
+    def get_feedback_dismissed_by(self, activity):
+        return [c.user_id for c in activity.activityparticipant_set.all() if c.feedback_dismissed]
```
nicksellen, line 194: Maybe worth a comment to explain why the filtering is being done in python, rather than the db, we can borrow the comment used elsewhere:

_we are filtering in python to make use of prefetched data_

nicksellen (changes requested):
Really nice work! :rocket: 

Especially given I think you said this is a whole new tech stack _including_ the language!? It looks really nice idiomatic python/karrot code, and I hope it was also fun learning some new tech along the way!

As in the frontend PR, just minor comments/suggestions really, then I think this is good to go!

There is a _bonus_ feature that you could consider: for some groups they _require_ feedback to be given within their group (e.g. Warsaw), and this could be made configurable on the level of the _activity type_ (which currently has the option whether to take feedback or not, but could be extended to define whether feedback is _required_ or optional). 

However, I'm very happy with this PR without that, and it probably makes more sense to do in another PR anyway...

karrot/subscriptions/receivers.py
nicksellen: tl;dr: nothing to see here! just Nick thinking too much :thinking: 

I was thinking that this still doesn't handle the case of leaving an activity and updating the count... but you can never leave an activity that it's possible to give feedback to (leaving is only before the activity, and feedback is only after the activity), so I think it's fine. There could be some nuance for the _during_ the activity time, we have some issues with that already, but yeah, don't need to think about them here.

nicksellen (approved):

nicksellen:
Woo! :partying_face: I approved!

I'll let you press the merge button for the nice warm feeling it might give you inside :) ... also we can get this one merged/deployed before dealing with the frontend.

(Once merged this will auto deploy the backend of dev.karrot.world).

PR #1125, opened 2021-03-24
ics export endpoint

amengsk:
this is for issue https://github.com/yunity/karrot-frontend/issues/2320

it adds the ics export endpoint to the server


karrot/activities/api.py
```diff
@@ -252,3 +254,22 @@ def remove(self, request, pk=None):
...
+
+    def get_queryset(self):
+        return self.queryset.filter(place__group__members=self.request.user)
+
+    def finalize_response(self, request, response, *args, **kwargs):
+        """specify a filename for the downloaded file"""
+        super(ActivityICSViewSet, self).finalize_response(request, response, *args, **kwargs)
+        if isinstance(response, Response) and response.accepted_renderer.format == 'ics':
+            response['content-disposition'] = 'attachment; filename=invite.ics'
+        return response
```
nicksellen, line 275: Instead of a new viewset (which then has it's own url prefix `/activities-ics/`), I think this could be neater as a detail method on the existing viewset, something like:

```py
    @action(
        detail=True,
        methods=['GET'],
        renderer_classes=(ICSCalendarRenderer, ),
        serializer_class=ActivityICSSerializer,
    )
    def ics(self, request, pk=None):
        response = self.retrieve(request)
        response['content-disposition'] = 'attachment; filename=invite.ics'
        return response
```

What do you think? 

(I'd also suggest making the filename unique to the activity, in the simplest way `activity-{id}.ics` perhaps? Or `{type}-{id}.ics`?)

karrot/activities/renderers.py
```diff
@@ -0,0 +1,44 @@
...
+        lines = [
+            'BEGIN:VEVENT',
+        ]
+        lines += [
+            '{}:{}'.format(key.upper().replace('_', '-'), value.replace('\n', '\\n')) for key, value in vevent.items()
+        ]
+        lines += [
+            'END:VEVENT',
+        ]
+        return '\r\n'.join(lines)
```
nicksellen, line 44: What do you think about using a library to handle the ics formatting? https://icspy.readthedocs.io/en/stable/ looks reasonable? Is there a specific reason to have handrolled rendering?

karrot/activities/tests/test_renderers.py
```diff
@@ -0,0 +1,31 @@
...
+            'DTSTART': '20210319T170000Z',
+            'DTEND': '20210319T180000Z',
+            'DESCRIPTION': 'Hello\nWorld!',
+            'TRANSP': 'OPAQUE'
+        },
+                                serializer=None)
+        self.expected_output = (
+            "BEGIN:VCALENDAR\r\n" + "VERSION:2.0\r\n" + "PRODID:-//Karrot//EN\r\n" + "BEGIN:VEVENT\r\n" +
+            "DTSTART:20210319T170000Z\r\n" + "DTEND:20210319T180000Z\r\n" + "DESCRIPTION:Hello\\nWorld!\r\n" +
+            "TRANSP:OPAQUE\r\n" + "END:VEVENT\r\n" + "END:VCALENDAR"
```
nicksellen, line 20: I guess yapf formatted this a bit oddly! Maybe a list of values joined by `\r\n` would be easier to read?

nicksellen (commented):
Hey! Thanks so much for the new ics export feature :star: 

I left a few comments for discussion, interested to hear what you think :)

nicksellen:
As a sidenote, I found it odd that my android phone didn't offer to add it into my calendar automatically when I downloaded the .ics, it seems the support is not as good as I had hoped (I use etar calendar, which has ics import, but seemingly not a handler for downloaded ics files, it attempts to open it in markor editor... not useful phone!)

karrot/activities/renderers.py
amengsk: its a simple format and i wasnt sure how happy you are with new deps. but i was considering https://pypi.org/project/icalendar/ actually. also not sure if its natural to make it work with drf nice? what architecture would you use?

karrot/activities/api.py
amengsk: sure why not. with a viewset i was thinking you could then also expose ics files for all pickups in a place, in a group, joined by a user… so not just a 'retrieve' view, but also a list view with filters etc. cause ics files can contain many events! but its not supported atm and maybe we can migrate to a viewset later if need be? i guess its no big deal if urls change

nicksellen: Yes, I thought about that too, it _looks like_ you can also do these `@action` things with `detail=False` and it would create an endpoint like `/activities/ics/` that comes 1 higher priority than `/activities/{id}/` so I _think_ that would work... didn't try it though.

I like the idea though, and it could be used with https://icsx5.bitfire.at/ nicely. For that subscription case though we'd need to think about the authentication aspect too... maybe it'd need a seperate viewset for that, or whatever. We can think of that if someone is motivated to build it :)

karrot/activities/renderers.py
nicksellen: In general I'm happier with libraries (maintained by somebody else) than hand-rolled stuff (have to somehow keep it my head), and the one you mentioned looks good too (more recently released as well).

Looks like it would be easy to use with your new activity serializer too... I think you could just replace the code inside your custom renderer with a call to the library. Does it make sense?

karrot/activities/tests/test_renderers.py
```diff
@@ -0,0 +1,40 @@
...
+            'transp': 'OPAQUE',
+            'attendee': attendee,
+            'toskip': None
+        },
+                                serializer=None)
+        self.expected_output = '\r\n'.join([
+            "BEGIN:VCALENDAR", "VERSION:2.0", "PRODID:-//Karrot//EN", "BEGIN:VEVENT",
+            "DTSTART;VALUE=DATE-TIME:20210319T170000Z", "DTEND;VALUE=DATE-TIME:20210319T180000Z",
+            "ATTENDEE;CN=\"Mary Smith\";ROLE=REQ-PARTICIPANT:MAILTO:marysmith@example.com",
+            "DESCRIPTION:Hello\\nWorld!", "TRANSP:OPAQUE", "END:VEVENT", "END:VCALENDAR"
```
nicksellen, line 29: minor: yapf will do the formatting nicely here but only if there is a trailing comma

nicksellen (approved):
Great! :superhero: Looks good to me, just one minor/optional point, but ready to merge otherwise! (I guess I need to press that as you're not in the "karrot-dev" team?)

nicksellen:
Merged! It'll get deployed to dev.karrot.world automatically now :rocket:

PR #1128, opened 2021-04-05
HINT: exclude in setup.py works on a per-package basis, not per-file

inktrap:
```
        exclude=['config/local_settings'],  # doesn't work :(
```

it is not working because config/local_settings is not a package but a *file* in a package; if it were a subpackage in config, then ``config.local_settings`` would work.
See: https://github.com/pypa/setuptools/blob/a4dbe3457d89cf67ee3aa571fdb149e6eb544e88/setuptools/__init__.py#L89
I would suggest to include a generic config.local-settings file when the package is built (isn't that the case anyway?) and the package should not be built on a local (dev) machine anyways, right?

To have the intended effect i added the exclusion of the data file for the config package, see the second commit or https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html (I didn't test this).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,12 +9,13 @@ setup(
     author_email='info@karrot.world',
     license='AGPL',
     packages=find_packages(
-        include=['config', 'karrot', 'karrot.*']
+        include=['config', 'karrot', 'karrot.*'],
     ),
     package_data={
         'config': ['options.env'],
         'karrot': ['*/templates/*.jinja2', 'COMMIT'],
     },
     include_package_data=True,
+    exclude_package_data={"config": ["local_settings"]},
     zip_safe=False
 )
--- a/setup.py
+++ b/setup.py
@@ -9,8 +9,7 @@ setup(
     author_email='info@karrot.world',
     license='AGPL',
     packages=find_packages(
-        include=['config', 'karrot', 'karrot.*'],
-        exclude=['config/local_settings'],  # doesn't work :(
+        include=['config', 'karrot', 'karrot.*']
     ),
     package_data={
         'config': ['options.env'],


inktrap:
NOTE: Due to the way the distutils build process works, a data file that you include in your project and then stop including may be “orphaned” in your project’s build directories, requiring you to run setup.py clean --all to fully remove them. This may also be important for your users and contributors if they track intermediate revisions of your project using Subversion; be sure to let them know when you make changes that remove files from inclusion so they can run setup.py clean --all

https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html

nicksellen (approved):
Ah quite right, thanks!

I included this bit in the build script just incase somebody does try to build with a local_settings.py ... but the builds all happen on the ci, so should be fine generally...

https://github.com/yunity/karrot-backend/blob/b1a305ed55e42b8a5b4ed6d0ff9fa1263ce6771c/scripts/build#L5-L8

Nice to get a bit more understanding of how include/exclude works... it's still a bit odd having `MANIFEST.in` and `package_data` ... would welcome a cleanup there too if you ever feel like it! (might need some more rejigging to put things into "packages")

inktrap:
Actually for manifest.in there is only

```
graft karrot/static
```

which means to include everything, including subdirs, in karrot/static, see https://packaging.python.org/guides/using-manifest-in/
However, there is no ``karrot/static``? Are they added during the ci workflow? E.g. after they have been created during minification/joining? https://github.com/yunity/karrot-backend/tree/master/karrot

Then if the location is karrot/static … why not add that as package data **if the level ALWAYS stays the same**, e.g. ``static/css/foo.css`` and ``static/js/foo.js``?

```
    package_data={
        'config': ['options.env'],
        'karrot': ['*/templates/*.jinja2', 'COMMIT', 'static/*/*'],
    },
```
If not … it gets somehow messy and I would keep the manifest.in
https://stackoverflow.com/questions/27664504/how-to-add-package-data-recursively-in-python-setup-py

nicksellen:
Yup, `karrot/static` gets created here during build https://github.com/yunity/karrot-backend/blob/e446467d6543d728c84cdea8b51142bc0b5b7ae9/scripts/build#L20

Potentially it could be many levels, it's essentially arbitrary files, here's the listing when I run collectstatic locally: https://gist.github.com/nicksellen/866a79a661410c9c3dc9dc30ee3c70b7

Seems like keeping manifest.in is better, I would have liked to put more stuff in there, but it didn't seem to work :/

inktrap:
I guess I would leave it like it is for now

PR #1132, opened 2021-04-23
support for ics subscriptions

amengsk:
quick prototype for https://github.com/yunity/karrot-frontend/issues/2334

you can try it out without the frontend, by subscribing http://localhost:8080/api/activities/ics/ to with your calendar app or by downloading it with your browser (both should work, with different auth methods).

@nicksellen i used a separate viewset again but if you know how to integrate it to the existing viewset, very happy to do so of course


nicksellen:
Hey! Thanks for this, I think could be really useful for people into this!

You're right that I would think reusing the existing activities view set would be preferable, as ics format is really just another representation of the same information the API already returns.

So, you could even see if it's possible/easy/clean to have something like a `?format=ics` parameter, and reuse even more of the existing code. I don't know if drf already provides a way to switch renderers/serializers based on query params?

Otherwise, I _think_ a non `detail` action should works, such that `/activities/ics/` would be resolved before `/activities/<id>/`... but I never tried these ones.

What do you think?

Also, which calendar client do you use?

amengsk:
> So, you could even see if it's possible/easy/clean to have something like a ?format=ics parameter, and reuse even more of the existing code. I don't know if drf already provides a way to switch renderers/serializers based on query params?

we can do that, but its not just the renderers and serializers that need to be adapted, you also need to disable pagination for ics and enable basic http auth.

we could also enable basic http auth for all methods in the activities API but then it changes some http statuses in the tests, it feels a bit weird to have to touch that

plus we might need to change the filtering capabilities too, if we want the ics views to support the filtering options that are currently implemented client-side (see https://github.com/yunity/karrot-frontend/issues/2334#issuecomment-822746441), although it's perhaps not too bad if they get enabled for the original activities API in the same go.

so i think it perhaps not so "clean" in the sense that its hijacking the nice declarative aspect of drf, but it can definitely be done

> Otherwise, I think a non detail action _should_ works, such that /activities/ics/ would be resolved before /activities/<id>/... but I never tried these ones.

the problem with that the action is either a detail or a list view, not both at the same time (the `action` decorator requires a `detail` parameter which is either true or false), so it seems hard to get `/activities/ics/` and `/activities/ics/<id>/` to be supported at the same time (you would need to define two methods with the same name and decorated differently).

nicksellen:
Yup, I see the dilemma for the `?format=ics` idea! Thanks for looking into it. Better to stick with what is easy for drf.

>  it seems hard to get /activities/ics/ and /activities/ics/<id>/ to be supported at the same time (you would need to define two methods with the same name and decorated differently)

This one seems more achievable though, as the `@action` annotation accepts a `url_path`  option:

> :param url_path: Define the URL segment for this action. Defaults to the name of the method decorated.

I do still get a niggly feeling from the API/REST perspective of using a totally separate viewset, but can see the benefit from the code side (as it has potentially different options/patterns to the REST version given it's a different use case).

How about this, I have expressed by vague thoughts, but will leave it to you to decide the approach to use, whether it's a separate viewset or multiple detail methods, or whatever else :)

amengsk:
> This one seems more achievable though, as the @action annotation accepts a url_path option.

ah good point. then i will try that

amengsk:
sorry for the mixup!

i have merged things in the same viewset, it works pretty well i would say

i tested the subscription feature with Evolution and Lightning, it works well for both

let me know what you think

amengsk:
checkout the frontend pr: https://github.com/yunity/karrot-frontend/pull/2368

nicksellen (approved):
Nice! I like that it keeps the existing single activity ics endpoint the same.

I think this is good to merge! Thanks :+1:

nicksellen:
It says it's out of date with master branch, do you want to rebase/merge that in? ... or I can press "squash and merge" here and hope GitHub does the right thing...

nicksellen:
It says it's out of date with master branch, do you want to rebase/merge that in? ... or I can press "squash and merge" here and hope GitHub does the right thing...

amengsk:
@nicksellen sure, now it should be good to squash and merge :)

PR #1133, opened 2021-04-28
Revoke Trust

pogopaule:
see https://github.com/yunity/karrot-frontend/pull/2352

### DIFF
--- a/karrot/groups/api.py
+++ b/karrot/groups/api.py
@@ -201,14 +201,7 @@ class GroupViewSet(
 
         return Response(data={})
 
-    @action(
-        detail=True,
-        methods=['POST'],
-        permission_classes=(IsAuthenticated, IsOtherUser),
-        url_name='revoke-trust',
-        url_path='users/(?P<user_id>[^/.]+)/revoketrust',
-        serializer_class=EmptySerializer
-    )
+    @trust_user.mapping.delete
     def revoke_trust(self, request, pk, user_id):
         """revoke trust for a user in a group"""
         self.check_permissions(request)
--- a/karrot/groups/api.py
+++ b/karrot/groups/api.py
@@ -203,6 +203,30 @@ class GroupViewSet(
 
     @action(
         detail=True,
+        methods=['POST'],
+        permission_classes=(IsAuthenticated, IsOtherUser),
+        url_name='revoke-trust',
+        url_path='users/(?P<user_id>[^/.]+)/revoketrust',
+        serializer_class=EmptySerializer
+    )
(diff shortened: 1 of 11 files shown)


karrot/groups/api.py
```diff
@@ -201,6 +201,30 @@ def trust_user(self, request, pk, user_id):
 
         return Response(data={})
 
+    @action(
+        detail=True,
+        methods=['POST'],
```
nicksellen, line 206: As mentioned in the frontend PR, using `DELETE` method on the same endpoint as the trust one would fit in better with what we've done so far.

nicksellen (commented):
Great, looks really good to me :100: 

... just the URL endpoint question open.

karrot/groups/api.py
pogopaule: done

karrot/groups/api.py
```diff
@@ -201,6 +201,23 @@ def trust_user(self, request, pk, user_id):
 
         return Response(data={})
 
+    @trust_user.mapping.delete
```
nicksellen, line 204: Ah, nice find using this [`mapping` thing](https://www.django-rest-framework.org/api-guide/viewsets/#routing-additional-http-methods-for-extra-actions), we could clean up other actions using that pattern one day! :+1:

nicksellen (approved):
Looks good to me now! :rocket:

pogopaule:
@nicksellen running `python manage.py makemessages` on my machine does not cause `karrot/locale/en/LC_MESSAGES/django.po` to change. But on circleci it does and fails the test pipeline...do you have an idea why?

nicksellen:
Have you run `./sync.py` to install latest dependencies? Even patch version differences can generate slightly different locale files, and I just merged an dependency update that was impacted in that way. I'm assuming you also just rebased on master.

pogopaule:
@nicksellen Thank you! `./sync.py` did the job!

karrot/history/models.py
```diff
@@ -43,6 +43,7 @@ class HistoryTypus(enum.Enum):
     ACTIVITY_TYPE_CREATE = 25
     ACTIVITY_TYPE_MODIFY = 26
     ACTIVITY_TYPE_DELETE = 27
+    USER_LOST_EDITOR_ROLE = 18
```
nicksellen, line 46: I just noticed this number is not unique, I guess was supposed to be `28` :)

pogopaule: Oh no 🤦 I'll fix that tomorrow in a new PR

PR #1134, opened 2021-05-04
Add country code and timezone to geo ip info

nicksellen:
To help https://github.com/yunity/karrot-frontend/issues/2355 be implemented this adds `country_code` field to geoip info we return. 

I also added a `timezone` field whilst I was there, which could perhaps be used to only load the client side timezone detection library if it's really needed (via a dynamic import). Just a thought.

This can be manually tested like this:

```
curl -s -H 'X-Forwarded-For: 191.249.32.2' localhost:8000/api/bootstrap/ | jq .geoip
{
  "lat": -19.8993,
  "lng": -43.957,
  "country_code": "BR",
  "timezone": "America/Sao_Paulo"
}
```

### DIFF
--- a/karrot/utils/geoip.py
+++ b/karrot/utils/geoip.py
@@ -31,8 +31,7 @@ def ip_to_city(ip):
     try:
         return geoip.city(ip)
     except AddressNotFoundError:
-        # we use "False" to mean we looked it up but couldn't find it
-        return False
+        return None
 
 
 @lru_cache()
@@ -42,5 +41,4 @@ def ip_to_lat_lon(ip):
     try:
         return geoip.lat_lon(ip)
     except AddressNotFoundError:
-        # we use "False" to mean we looked it up but couldn't find it
-        return False
+        return None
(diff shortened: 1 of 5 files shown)


karrot/utils/geoip.py
```diff
@@ -24,6 +24,17 @@ def get_client_ip(request):
         return request.META.get('REMOTE_ADDR')
 
 
+@lru_cache()
+def ip_to_city(ip):
```
pogopaule, line 28: Codecov complains about missing tests for the exception cases. Is that something that should be added?

pogopaule:
The code looks good to me and the curl works.

pogopaule (approved):
The code looks good to me and the curl works.

karrot/utils/geoip.py
nicksellen: I added some more tests for that now :) ... also better typing, by returning `None` not `False` when it can't find the address.

PR #1135, opened 2021-05-06
fix(HistoryTypus): set USER_LOST_EDITOR_ROLE to 28

pogopaule:


pogopaule:
Not sure if it is worth having that test...let me know what you think

karrot/history/tests/test_models.py
```diff
@@ -35,3 +35,12 @@ def test_activity_left_late(self):
         self.assertEqual(annotated.activity_leave_seconds, leave_seconds_before_activity)
         self.assertEqual(History.objects.activity_left_late(seconds=leave_seconds_before_activity).count(), 1)
         self.assertEqual(History.objects.activity_left_late(seconds=leave_seconds_before_activity - 1).count(), 0)
+
+
+class HistoryTypusTests(TestCase):
+    def test_typus_ids_are_unique(self):
+        list_of_all_history_typus = list(map(lambda tuple: tuple[1], HistoryTypus.__members__.items()))
```
nicksellen, line 42: I think the list comprehension is more idiomatic (and neater) in python, rather than map + lamda, and also `enum.Enum` is iterable, so it can become:
```py
list_of_all_history_typus = [item.value for item in HistoryTypus]
```

karrot/history/tests/test_models.py
```diff
@@ -35,3 +35,12 @@ def test_activity_left_late(self):
         self.assertEqual(annotated.activity_leave_seconds, leave_seconds_before_activity)
         self.assertEqual(History.objects.activity_left_late(seconds=leave_seconds_before_activity).count(), 1)
         self.assertEqual(History.objects.activity_left_late(seconds=leave_seconds_before_activity - 1).count(), 0)
+
+
+class HistoryTypusTests(TestCase):
+    def test_typus_ids_are_unique(self):
```
nicksellen, line 41: +1 for test :) ... I thought about it before, it's an easy mistake to make, and I'm sure has happened before...

nicksellen (approved):
Great, I'm all in for the test.

I made an optional python-coding suggestion, so I leave that for you :)

karrot/history/tests/test_models.py
pogopaule: Yes, that would be a lot more readable. Unfortunately this does not work in the tests since the idiomatic approach kicks out the duplicate values :-/

pogopaule: Just realized I can use list comprehension anyways. Looks a lot better now, what do you think?

nicksellen: Ah! Gotcha, maybe just include a comment to that effect, to save the next person having the same thought I did :)

So, using that idea, something like this seems to work too:
```py
# some comment about why... 
len(HistoryTypus) != len(HistoryTypus.__members__)
```

pogopaule: Nice! Can not get shorter than that ;) Made a new commit.

PR #1137, opened 2021-05-08
Fix place statistics, weight avg -> sum

nicksellen:
A while back we changed feedback weight to always be summed if multiple people give it, but it seems we didn't change the calculation for the place statistics...


nicksellen:
Approved by @tiltec here --> https://community.foodsaving.world/t/wrong-sum-up-of-pickup/711/4

PR #1139, opened 2021-05-10
Add more activity statistics

nicksellen:
Closes https://github.com/yunity/karrot-frontend/issues/2336 (well, most of it I think). Also see https://community.foodsaving.world/t/more-statistics-display/489/8.

This adds:
- **missed activities** - only relevant if there is no user filter, as a single user can't miss an activity by definition
- **feedback count** - in addition to feedback weight that it already returned
- refined the **left** / **left late** concept, so now:
  - **left** = activities left at any time, which were maybe later fulfilled, or not
  - **left late** = same as above, but left within 24 hours of start
  - **left missed** (_new!_) = activities left at any time, and later were missed (nobody did them)
  - **left late missed** (_new!_) = same as above, but left within 24 hours of start
- filter by **activity type**

Before this PR, both the **left** and **left late** would check that the activities were later _not full_ (rather than missed entirely). I wasn't sure at the time which was better.

With the new way the frontend can calculate the "missed but later fulfilled" value too if needed (`left - left missed`).

I also fixed a bug that I _think_ was double counting stats for activities with multiple feedbacks... I added `distinct` to all the `Count()` calls to resolve that.

### DIFF
--- a/karrot/stats/api.py
+++ b/karrot/stats/api.py
@@ -62,11 +62,15 @@ class ActivityHistoryStatsViewSet(ListModelMixin, GenericViewSet):
             .values('place', 'group') \
             .filter(typus__in=[
                 HistoryTypus.ACTIVITY_DONE,
+                HistoryTypus.ACTIVITY_MISSED,
                 HistoryTypus.ACTIVITY_LEAVE,
             ]) \
             .annotate(
                 done_count=Count('activity', filter=Q(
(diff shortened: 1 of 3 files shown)


nicksellen:
@tiltec [said](https://chat.karrot.world/channel/karrot-dev?msg=yJP3wxYYxktjCwfyi):

> I had a look, especially on the backend part, and felt a bit out of depth there.. but calculations looked correct! 

So merging :)

PR #1141, opened 2021-05-31
Message list: split queries to avoid slow cases

tiltec:
Related to https://github.com/yunity/karrot-frontend/issues/2369

Moves annotation of replies_count and unread_replies_count into a separate query, to avoid extremely slow cases.

I tested `/api/messages/?conversation=42199` on my machine with the karrot.world production db:
- 95th percentile before: multiple seconds
- 95th percentile after: 460ms

### DIFF
--- a/karrot/conversations/api.py
+++ b/karrot/conversations/api.py
@@ -260,7 +260,6 @@ class ConversationMessageFilter(filters.FilterSet):
 
 class ConversationMessageViewSet(
         mixins.CreateModelMixin,
-        mixins.ListModelMixin,
         mixins.RetrieveModelMixin,
         PartialUpdateModelMixin,
         GenericViewSet,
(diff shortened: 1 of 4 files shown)


karrot/conversations/api.py
```diff
@@ -300,14 +300,28 @@ def get_queryset(self):
...
-
         if self.request.query_params.get('thread', None):
             return qs.only_threads_and_replies()
 
         return qs.exclude_replies()
 
+    def list(self, request, *args, **kwargs):
+        # Workaround to avoid extremely slow cases: split up query
+        # 1. get message ids, including costly access control
+        # 2. get data, including costly annotations
```
nicksellen, line 311: Maybe include a link to the issue too, in case in the future it becomes untrue (query planner can improve), and then the trail will be possible to follow.

nicksellen (changes requested):
Great work! :star: That'll be really beneficial to a lot of users :smiling_face_with_three_hearts: :woman_dancing: :dancing_men: 

Looks like tests are still failing, some because of query count, but also some others where the diff is too long to see the detail.

tiltec:
I'm investigating failing tests, seems like an actual problem with my code that I don't understand yet.

Also I'll try to apply similar changes to `/api/messages/my_threads/`.

tiltec:
Should be good now!

nicksellen (approved):
Seems good! I didn't look in detail, but put trust in you and the tests, and we can fix in the future if anything else amiss.

It seems it should lead to a big improvement in life quality for karrot users, so good to roll out soon! I'm wanting to do a production deployment anyway soon, now that we have the new landing page ready.

Really appreciate the efforts here! :1st_place_medal:

tiltec:
Ooops, I didn't squash the commits :(

PR #1143, opened 2021-06-02
Add request and db timeout options

nicksellen:
Relating to https://github.com/yunity/karrot-frontend/issues/2369

These timeouts are to improve handling the situation where they are excessively slow db queries:
1. a db statement timeout, this means postgres will abort the query when it reaches that limit
2. daphne request timeout, so the webserver will abort the request

I set the defaults both to 60 seconds, but might be a good idea to set them to shorter values in production, as 60s is pretty unusable...

The idea is to avoid blocking the server processes and database connections with long running queries.

### DIFF
--- a/config/asgi_app.py
+++ b/config/asgi_app.py
@@ -2,12 +2,25 @@ from channels.auth import AuthMiddlewareStack
 from channels.routing import ProtocolTypeRouter
 from django.conf import settings
 from django.core.asgi import get_asgi_application
+from django.db.backends.signals import connection_created
+from django.dispatch import receiver
 from starlette.responses import Response
 
 from starlette.staticfiles import StaticFiles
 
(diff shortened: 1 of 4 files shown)


tiltec (approved):

tiltec:
As mentioned in rocketchat: I wonder how it works with some background tasks, I think the weekly email has some slower queries that are unproblematic (and shouldn't be killed)

tiltec:
What about stats/logs? Would be nice to have some data in influxdb which requests and queries get killed...

nicksellen:
> As mentioned in rocketchat: I wonder how it works with some background tasks, I think the weekly email has some slower queries that are unproblematic (and shouldn't be killed)

As answered in rocketchat:  I set the db timeouts to only be for web requests, for that reason 

> What about stats/logs? Would be nice to have some data in influxdb which requests and queries get killed...

yes! I imagine they will end up in sentry, it ends up as a 500 error back to the client, with a django-y error page and a postgres error message....

PR #1145, opened 2021-06-06
activity-list enpoint: avoid n+1 queries

tiltec:
Looking up `activity_type` caused n queries. The endpoint is used frequently, so it seems better to avoid this.

This should roughly halve mean response time from 400ms to 200ms. For groups with lots of activities, response time was around 700ms - there's no pagination (yet).

I applied the same optimization to `activity-retrieve`.


nicksellen (approved):
Great! Good idea 👍

PR #1146, opened 2021-06-08
vendor django-influxdb-metrics

tiltec:
We used a fork for a long time, so we might as well copy in relevant code to improve code visibility and maintainability.

Reason for the fork: removes some fast-changing stats from tags in order to not overload influxdb

Additional benefits:
- removed some unused code and settings
- removed 3 unused dependencies
- avoids Django deprecation warning about `is_ajax()`

`INFLUXDB_HOST_TAG` was only used for the few builtin stats, not for the large body of our custom stats. Therefore I decided to drop it. Can drop it from yuca ansible as well then.


nicksellen (approved):
Looks good! Not a very thorough review, but as it's mostly copying it in, I guess it should be fine :)

PR #1147, opened 2021-06-09
activity-ics: use simpler way to disable pagination

tiltec:
Just saw this in https://github.com/yunity/karrot-backend/pull/1132, thought this might a nicer way to disable pagination for a specific endpoint: set `pagination_class=None`.
@nicksellen @amengsk


nicksellen (approved):
On the assumption it works, seems good, and simpler!

PR #1148, opened 2021-06-10
Switch to drf-spectacular for API documentation

tiltec:
[django-rest-swagger](https://github.com/marcgibbons/django-rest-swagger) has been unmaintained for a long time and we rely on [our fork that contains a custom fix](https://github.com/tiltec/django-rest-swagger). Might be good to switch before some Django/DRF/Python version upgrade forces us to.

[drf-spectacular](https://github.com/tfranzel/drf-spectacular) seems the most up-to-date replacement. The alternative, drf-yasg, recommends using drf-spectacular for new projects because it uses OpenAPI 3.0 (https://github.com/axnsan12/drf-yasg).

Documentation seems nice, especially how to customize the generated schema: https://drf-spectacular.readthedocs.io/en/latest/customization.html

As before, I set it up to serve swagger-ui at `/docs/`. swagger-ui uses a new endpoint to get the schema at `/docs/schema/`.

image

The swagger-ui version is upgraded, looks similar but differs in many details. Notable differences:
- reloading respects previous position on the page, by using html anchors - very useful for fiddling with schema modifications
- long list of all schemas at the bottom, some with confusing names (`PatchedOffer`?) - EDIT: now collapsed by default!
- a bit slower to load
- ~~collapsible groups moved one level higher, which make them useless~~

I think it should be possible to fix the collapsing by having the root at `/api/` instead of `/`, I didn't check yet. EDIT: now collapsed!

Also, drf-spectacular gives useful warnings when running `manage.py spectacular --file schema.yaml --validate` on the command line, e.g. about cases where it couldn't identify the schema for some endpoint. Could be useful to run this in CI or pre-push hook to make/keep our documentation top-notch :)
UPDATE: with the test now retrieving `/docs/schema/`, these warnings show up during normal test runs.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -158,6 +158,13 @@ SPECTACULAR_SETTINGS = {
     'TITLE': 'Karrot API',
     'DESCRIPTION': '_could link to rocket chat or community forum here_',
     'VERSION': '0.1',
+    'SCHEMA_PATH_PREFIX': '/api/',
+    'SWAGGER_UI_FAVICON_HREF': '/favicon.ico',
+    'SWAGGER_UI_SETTINGS': {
+        'deepLinking': True,
+        'defaultModelsExpandDepth': 0,
+        'docExpansion': 'none',
+    },
 }
 
 MIDDLEWARE = (
--- a/config/settings.py
+++ b/config/settings.py
@@ -134,7 +134,7 @@ INSTALLED_APPS = (
     'corsheaders',
     'rest_framework',
     'rest_framework.authtoken',
-    'rest_framework_swagger',
+    'drf_spectacular',
     'anymail',
     'timezone_field',
     'django_jinja',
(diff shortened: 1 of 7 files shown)


nicksellen:
Oh, also regarding auth, the existing one has a "Session login" button:

sessionlogin

which redirects to DRF auth page at `/api-auth/login/?next=/docs/`:

drflogin

... which is the only way to login from that page (the "Authorize" button offers basic auth on the old one, and basic/session/token auth on the new one... none of which taken email/password).

tiltec:
I pushed some configuration changes for swagger ui which make it more behave like now: collapse all by default.

For auth, that's tricky indeed. django-rest-swagger uses some templating to add the header: https://github.com/marcgibbons/django-rest-swagger/blob/master/rest_framework_swagger/templates/rest_framework_swagger/index.html
We could [override drf-spectacular's template](https://docs.djangoproject.com/en/3.2/howto/overriding-templates/) with our own, it's [not doing much right now](https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/templates/drf_spectacular/swagger_ui.html).

The other way to log in: use the "try it out" feature on `/api/user/auth/` - should work fine, but doesn't support password managers and it's quite unintuitive.

Same goes for embedding swagger js, it _could_ live in our template folder. But updating it wouldn't be covered by pip, so it would mean extra work/complexity.

tiltec:
I did manage to overwrite the template to add session auth and hide the other auth button :tada: 

image

tiltec:
I'm happy to fix some of the warnings it shows about our schema, but possibly in another PR after this is merged?

nicksellen (approved):
Great, work, seems perfect! :+1:

PR #1150, opened 2021-06-13
Get rid of errors and warnings by drf-spectacular

tiltec:
Follow up to #1148, this PR adds some type annotations and makes changes where necessary to please drf-spectacular. 

Types of changes:
- add annotations, mostly to custom fields and SerializerMethodFields
- replace ReadOnlyField(default=...) with SerializerMethodFields, because the former seem impossible to annotate
- add serializer_class to viewsets that actually don't need them, to help drf-spectacular determine return types
- define some enum names couldn't be determined automatically (possibly because they are referenced from multiple endpoints)
- hardcode two `operation_id`s because they clashed

I also found out that we can simplify our enums if we use Django's new [`TextChoices`](https://docs.djangoproject.com/en/3.0/ref/models/fields/#enumeration-types) instead. It has a built-in way to produce `.choices`, so we can stop building them ourselves, as in `[(status.value, status.value) for status in ActivityTypeStatus]`

I sometimes used Python typings instead of `extend_schema` for simple cases, because they require less or no imports. Also, they seems easier to parse for humans, not having to worry about understanding another decorator.


karrot/groups/migrations/0027_auto_20180309_0936.py
```diff
@@ -17,6 +17,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='group',
             name='status',
-            field=models.CharField(choices=[('active', 'active'), ('inactive', 'inactive'), ('playground', 'playground')], default='active', max_length=100),
+            field=models.CharField(choices=[('active', 'Active'), ('inactive', 'Inactive'), ('playground', 'Playground')], default='active', max_length=100),
```
nicksellen, line 20: What is the significance of changing the casing here? (and for the activity type status)

karrot/groups/serializers.py
```diff
@@ -76,8 +81,17 @@ class GroupDetailSerializer(GroupBaseSerializer):
...
+
+    @staticmethod
+    def get_member_inactive_after_days(_) -> int:
+        return settings.NUMBER_OF_DAYS_UNTIL_INACTIVE_IN_GROUP
+
+    issue_voting_duration_days = serializers.SerializerMethodField()
+
+    @staticmethod
+    def get_issue_voting_duration_days(_) -> int:
+        return settings.VOTING_DURATION_DAYS
```
nicksellen, line 94: It's a shame there isn't a typed version of the `ReadOnlyField` to avoid these methods :/

karrot/userauth/api.py
```diff
@@ -115,6 +119,7 @@ def post(self, request):
 
 class ResendMailVerificationCodeView(views.APIView):
     permission_classes = (IsAuthenticated, MailIsNotVerified)
+    serializer_class = inline_serializer(name='', fields={})  # for OpenAPI generation with drf-spectacular
```
nicksellen, line 122: Maybe as the `inline_serializer(name='', fields={})` snippet is used three times, it could be extracted and given a meaningful name? Reflecting that it's basically a "empty serializer" as I understand it?

nicksellen (approved):
Great! I like having some more nudges towards typing, even if it's primarily to make drf-spectacular happier in this case. I had looked into the python typing stuff more widely at various points.

I added a few non-blocking thoughts/comments.

karrot/groups/migrations/0027_auto_20180309_0936.py
tiltec: First value is the actual one, second one is the "human-readable" one, e.g. for auto-generated dropdown fields in django admin interface. I believe this change shouldn't affect us, it's just a side effect from switching from `Enum`s to Django's `TextChoices`.

karrot/groups/serializers.py
tiltec: I think drf-spectacular is able to derive typing from `ReadOnlyField´ if the `source` parameter is given; our use case is a bit special by just providing the `default` value.

karrot/userauth/api.py
tiltec: Funnily, we already have a proper `EmptySerializer` somewhere in our code, maybe that one can be used instead.

PR #1151, opened 2021-06-14
GroupPreviewSerializer: remove member list and add annotations to improve performance

tiltec:
`/api/groups-info/` and `/api/bootstrap/` seem to be quite slow (average >400ms) and are called for every app load, so it seems useful to optimize. (https://grafana.karrot.world/d/000000003/requests?orgId=3)

So far, we always delivered a complete list of member ids for every group, because it seemed straightforward and the frontend relies on it. But we already prepared for its removal, by adding `member_count` and `is_member` to GroupPreviewSerializer in #1072.

The endpoint seems slow for two reasons:
1. we always send info about all groups, about 200 on karrot.world (most of them inactive)
2. calculation is done in the serializer, which takes most of the time (as profiled by Silk: database queries are fast)

This PR covers the second issue by removing the member list together with the prefetch query, and adding annotations to shift the work into the database.

This needs frontend support before it can get merged: https://github.com/yunity/karrot-frontend/pull/2383

### DIFF
--- a/karrot/groups/serializers.py
+++ b/karrot/groups/serializers.py
@@ -370,9 +370,8 @@ class GroupPreviewSerializer(GroupBaseSerializer):
         return group.get_application_questions_or_default()
 
     def get_member_count(self, group) -> int:
-        annotated = getattr(group, 'member_count', None)
-        if annotated is not None:
-            return annotated
+        if hasattr(group, 'member_count'):
+            return group.member_count
 
         return group.members.count()
 
@@ -381,9 +380,8 @@ class GroupPreviewSerializer(GroupBaseSerializer):
         if not user or user.is_anonymous:
             return False
 
-        annotated = getattr(group, 'is_user_member', None)
-        if annotated is not None:
-            return annotated
+        if hasattr(group, 'is_user_member'):
+            return group.is_user_member
 
         return user in group.members.all()
 
--- a/karrot/groups/serializers.py
+++ b/karrot/groups/serializers.py
@@ -341,7 +341,6 @@ class GroupPreviewSerializer(GroupBaseSerializer):
             'address',
             'latitude',
             'longitude',
-            'members',
             'member_count',
             'is_member',
             'status',
@@ -355,11 +354,22 @@ class GroupPreviewSerializer(GroupBaseSerializer):
         return group.get_application_questions_or_default()
 
     def get_member_count(self, group):
+        annotated = getattr(group, 'member_count', None)
+        if annotated is not None:
+            return annotated
+
         return group.members.count()
 
     def get_is_member(self, group):
         user = self.context['request'].user if 'request' in self.context else None
-        return user in group.members.all() if user else False
+        if not user or user.is_anonymous:
+            return False
+
+        annotated = getattr(group, 'is_user_member', None)
+        if annotated is not None:
+            return annotated
+
+        return user in group.members.all()
 
 
 class GroupJoinSerializer(GroupBaseSerializer):
--- a/karrot/offers/tests/test_api.py
+++ b/karrot/offers/tests/test_api.py
@@ -1,7 +1,7 @@
 from unittest.mock import patch
 
 from rest_framework import status
-from rest_framework.test import APITestCase, APITransactionTestCase
+from rest_framework.test import APITestCase
 
 from karrot.groups.factories import GroupFactory
 from karrot.offers.factories import OfferFactory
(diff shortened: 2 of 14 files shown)


nicksellen:
I'll wait for reviewing this until it's clearer what https://github.com/yunity/karrot-frontend/pull/2383 might need from the backend.

karrot/offers/tests/test_api.py
```diff
@@ -181,7 +181,8 @@ def test_create_offer(self, prepare_email):
                     'image': image_file
                 }],
             }
-            response = self.client.post('/api/offers/', data=encode_data_with_images(data))
+            with self.captureOnCommitCallbacks(execute=True):
```
tiltec, line 184: Django 3.2 allows us to use this instead of `TransactionTestCase`, should speed up test run!

karrot/groups/serializers.py
```diff
@@ -371,11 +370,22 @@ def get_application_questions(self, group) -> str:
...
         user = self.context['request'].user if 'request' in self.context else None
-        return user in group.members.all() if user else False
+        if not user or user.is_anonymous:
+            return False
+
+        annotated = getattr(group, 'is_user_member', None)
+        if annotated is not None:
+            return annotated
+
+        return user in group.members.all()
```
tiltec, line 388: I wondered if this would be easier to read:

```py
try:
  return group.is_user_member
except AttributeError:
  return user in group.members.all()
```

karrot/subscriptions/receivers.py
```diff
@@ -211,8 +211,8 @@ def send_group_detail(group, user=None):
 
 
 def send_group_preview(group):
-    preview_payload = GroupPreviewSerializer(group).data
     for subscription in ChannelSubscription.objects.recent():
+        preview_payload = GroupPreviewSerializer(group, context={'request': MockRequest(user=subscription.user)}).data
```
tiltec, line 215: GroupPreviewSerializer needs request user to work properly, this was broken before but uncritical (delivered invalid `isMember` fields via websocket).

tiltec:
This should be ready for review! @nicksellen 

I made a few slightly related changes:
- replace all usages of `TransactionTestCase` with `TestCase` and `captureOnCommitCallbacks` for consistency and to maybe speed up tests (didn't measure)
- added more assertions to existing webhook tests

karrot/groups/serializers.py
nicksellen: I find this way cleaner looking (and also used elsewhere):

```py
if hasattr(group, 'is_user_member'):
  return group.is_user_member

return user in group.members.all()
```

karrot/offers/tests/test_api.py
nicksellen: Nice! Seems a good find :+1:

karrot/subscriptions/receivers.py
nicksellen: Ah, good catch! :star:

karrot/subscriptions/receivers.py
```diff
@@ -244,15 +244,44 @@ def send_group_membership_updates(sender, instance, created, **kwargs):
         # notification types are only visible to one user
         send_group_detail(group, user=membership.user)
 
+
+@receiver(post_save, sender=GroupMembership)
+@on_transaction_commit
```
nicksellen, line 249: Is it needed to add this?

karrot/subscriptions/receivers.py
```diff
@@ -244,15 +244,44 @@ def send_group_membership_updates(sender, instance, created, **kwargs):
...
+                topic='groups:user_joined',
+                payload={
+                    'user_id': membership.user_id,
+                    'group_id': group.id,
+                }
+            )
 
 
 @receiver(post_delete, sender=GroupMembership)
+@on_transaction_commit
```
nicksellen, line 268: As before, is this needed?

nicksellen (approved):
Great! I really appreciate seeing these performance-improving things coming in :bullettrain_front: 

I'll approve, and let you decide what to do with my comments :)

karrot/subscriptions/receivers.py
tiltec: It seems conceptually necessary to prevent a race condition - adding members is part of a larger transaction (usually accepting an application), and the frontend would try to reload the user list immediately after receiving the websocket event. If the frontend reloads before the transaction commits, it would receive the outdated list.

tiltec: It seems conceptually necessary to prevent a race condition - adding members is part of a larger transaction (usually accepting an application), and the frontend would try to reload the user list immediately after receiving the websocket event. If the frontend reloads before the transaction commits, it would receive the outdated list.

With this, `application.accept()` is already wrapped in `@transaction.atomic`, so it should only send out the websocket event after the transaction has been committed.

nicksellen: Yup, seems good. I actually wondered if it's best to _always_ have the receivers `@on_transaction_commit`. I think at some point I played around with a custom annotation to do this, and also run them in huey jobs, as some potentially are slow. Anyway, that's another story :)

PR #1155, opened 2021-06-23
FCM: prevent cleanup and stats if response seems invalid

tiltec:
Adding a quick workaround because I couldn't reproduce the problem.

Related to https://chat.karrot.world/channel/karrot-dev?msg=uTdC7GHhAhi3dZ42k


nicksellen (approved):
Yup, seems good enough to try, can keep an eye on it in sentry...

PR #1156, opened 2021-06-24
websocket updates: provide empty request.META for GroupPreviewSerializer

tiltec:
By providing a mock `request` object for websocket updates, we got the error of missing `request.META` that is used by `DistanceField`. This PR provides an empty `META` dict as workaround.

Fixes https://sentry.io/organizations/foodsaving-worldwide/issues/2474312378

---

But now that I think about it, we actually might need a proper solution here. This change would set `distance` to `None` when the group update is delivered via websocket, which will mix up the sorting of group gallery.

The problem likely existed before in production, as `request` wasn't set for GroupPreviewSerializer in websocket updates. This PR would restore previous buggy behavior. Currently in production, people would not get group updates via websocket at all.

But maybe it would be better to go in the direction @nicksellen mentioned before, by just telling the frontend "groups have changed" and the frontend will reload them as it pleases. As this is a proper request, it would take the "normal" path with a full request object available.

### DIFF
--- a/karrot/groups/serializers.py
+++ b/karrot/groups/serializers.py
@@ -293,7 +293,7 @@ class DistanceField(Field):
     It may return None under various conditions:
     - the GeoIP2 libary was not initialized (missing the files)
     - there is no request context
-    - we cannot detirmine the IP address of the client
+    - we cannot determine the IP address of the client
     - the IP address cannot be found in the database
     """
     def __init__(self, **kwargs):
--- a/karrot/subscriptions/tests/test_receivers.py
+++ b/karrot/subscriptions/tests/test_receivers.py
@@ -463,9 +463,10 @@ class GroupReceiverTests(WSTestCase):
         super().setUp()
         self.member = UserFactory()
         self.user = UserFactory()
-        self.group = GroupFactory(members=[self.member])
+        self.group = GroupFactory(members=[self.member], latitude=48.0, longitude=12.0)
 
-    def test_receive_group_changes(self):
+    @patch('karrot.utils.geoip.geoip')
+    def test_receive_group_changes(self, geoip):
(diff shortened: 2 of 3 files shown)


nicksellen:
I wonder if we can get hold of the IP address for the websocket connection during the initial connection, then store it somewhere that we can use it for this `META` object?

nicksellen:
We can access the headers with `self.scope['headers']` in the websocket consumer, but not directly in the receiver (as the receiver might be running in a different process), but we could store the ip address in the `ChannelSubscription` when the user connects https://github.com/yunity/karrot-backend/blob/75d1db79afd78ba767c14203c08ab581fb16e1fe/karrot/subscriptions/consumers.py#L58-L60

Not sure which value from the header, might need to implement the logic to check through some headers, e.g. `x-forwarded-for`...

Seems it should work nicely, and be less work to implement than changing the data access patterns in the frontend too...

nicksellen:
It seems this issue is preventing trust being given at the moment for various people/groups.

See https://sentry.io/share/issue/1515f95f6835446b98d63e785cdae23b/

Sending the trust updates results in the group previews being sent, which is always using the `MockRequest`.

I guess at the moment giving trust only works when there are no users to push the update to :/ ?

tiltec:
There's also `self.scope['client']`, which contains the remote host ip and port: https://github.com/django/asgiref/blob/main/specs/www.rst

Not sure how it works through proxies, but websockets might behave differently than plain HTTP requests.

---

In any case, we could merge and push this PR as a hotfix for giving trust carrots whilst we are working on a proper solution!

nicksellen (approved):
Approved for quick fix for current issues :)

(Still worth finding a solution for the more nuanced niggles later)

PR #1161, opened 2021-07-02
websockets: use client ip for distance calculation

tiltec:
Follow-up to https://github.com/yunity/karrot-backend/pull/1156

This PR adds a `client_ip` field to `ChannelSubscription` and fills it from `self.scope['client']` during websocket connect. For now, it is only used to calculate the distance field in group preview updates.

I used a `GenericIPAddressField`, which seems semantically correct. But it does some normalization, so in theory it's possible that the distance calculation differs from using the non-normalized IP address - not sure how flexible the GeoIP thing is. Could also make it into a `TextField` to avoid this possibility. 

I made two (un)related changes:
- added a cronjob to clean up old channel subscriptions, and set the timeout from 5 to 30 minutes because I don't know how long connections can actually live
- cleared the `lru_cache` during testing, because some old tests started failing when I added a new one


karrot/subscriptions/consumers.py
```diff
@@ -57,7 +57,17 @@ def connect(self):
         if 'user' in self.scope:
             user = self.scope['user']
             if not user.is_anonymous:
-                ChannelSubscription.objects.create(user=user, reply_channel=self.channel_name)
+                # scope['client'] may contain client ip address
+                # https://github.com/django/asgiref/blob/e8797d47f4006e4e817fda507ec1eacfed4faf60/specs/www.rst
```
nicksellen, line 61: I wonder if this includes x-forwarded-for stuff? Might just be local ip (which would just be internal one on the server).

I had a quick look at both daphne and uvicorn and they both seem to support x-forwarded-for for websockets as a far as I could tell. We could check once it's deployed to dev...

karrot/subscriptions/receivers.py
```diff
@@ -212,7 +212,12 @@ def send_group_detail(group, user=None):
 
 def send_group_preview(group):
     for subscription in ChannelSubscription.objects.recent():
-        preview_payload = GroupPreviewSerializer(group, context={'request': MockRequest(user=subscription.user)}).data
+        preview_payload = GroupPreviewSerializer(
+            group,
+            context={
+                'request': MockRequest(user=subscription.user, META={'REMOTE_ADDR': subscription.client_ip})
+            }
+        ).data
```
nicksellen, line 220: There are other places that use a `MockRequest` in the context of a subscription, maybe worth a `subscription -> MockRequest` function, and using it in the other places too?

karrot/subscriptions/tasks.py
```diff
@@ -196,3 +198,12 @@ def notify_subscribers_by_device(subscriptions, *, click_action, fcm_options):
             'click_action': click_action,
         }
     )
+
+
+@db_periodic_task(crontab(hour='*/24', minute=35))  # every 24 hours
+def delete_old_channel_subscriptions():
+    with timer() as t:
+        # delete old channel subscriptions after some minutes of inactivity
+        ChannelSubscription.objects.old().delete()
```
nicksellen, line 207: Oh, we must have a lot of these by now!

nicksellen (approved):
Great! I like it. I'll approve, but also had a suggestion for a little bit more work if you feel like it :hammer:

karrot/subscriptions/consumers.py
tiltec: I checked and it does resolve to the actual client ip, so it seems to include headers. but let's double-check that once deployed to dev!

karrot/subscriptions/receivers.py
tiltec: Yup, good idea! I'll remember this for another PR :smile:

PR #1162, opened 2021-07-03
replace raven with sentry_sdk

tiltec:
`raven` has been marked as legacy for a while, so I thought it might be time to switch to `sentry_sdk`.

I should test this somehow, but it seems a bit cumbersome. Maybe I can just push to dev and check things there...?

I'm uncertain about Django's `LOGGING` configuration, not sure which parts are still needed and if they still work after the upgrade.

References:
- django logging docs: https://docs.djangoproject.com/en/3.2/topics/logging/
- django default logging conf: https://github.com/django/django/blob/main/django/utils/log.py
- sentry_sdk docs: https://docs.sentry.io/platforms/python/
- legacy raven client: https://github.com/getsentry/raven-python


karrot/groups/tasks.py
```diff
@@ -1,16 +1,15 @@
-from anymail.exceptions import AnymailAPIError
 from datetime import timedelta
```
nicksellen, line 1: How come you removed this line? Seems unrelated?

```py
stats_utils.periodic_task('group__mark_inactive_groups', seconds=t.elapsed_seconds)
```

(GitHub wouldn't let me comment on the actual removed line...)

nicksellen (approved):
Looks good! Yup, let's push to dev and test there.

Seems a random removed line? I approved, but maybe have a look at that bit before merging.

karrot/groups/tasks.py
tiltec: Not sure why this shows up, I didn't change anything about this line. It's actually the last line in the file, but it shows up as line number zero? Might be a problem with github...

PR #1166, opened 2021-07-13
Add env variable to set sentry environment

tiltec:
Should make it easier to differentiate between seperate deployments using the same DSN, as we have for karrot.world and dev.karrot.world.

I already made [this change](https://github.com/yunity/yuca/commit/2d5bd302165ebb01fd719a8a8e207db0c4b3f05f) in the yuca ansible repo to set SENTRY_ENVIRONMENT to `site`, which should be `karrot-world` and `karrot-dev` for our purposes.

### DIFF
--- a/config/options.env
+++ b/config/options.env
@@ -101,6 +101,7 @@ INFLUXDB_PASSWORD=
 #-----------------------------------------------
 
 SENTRY_DSN=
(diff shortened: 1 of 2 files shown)


nicksellen (approved):
Great, good idea! :bulb:

PR #1167, opened 2021-07-14
Add trust for role

nicksellen:
Forum thread: https://community.foodsaving.world/t/applicant-trial-pickup-proposal/575
Backend part for _trust for role_ feature https://github.com/yunity/karrot-frontend/issues/2361

It:
* adds `role` field to `Trust` model
* accepts `role` field in the trust user / revoke trust API endpoints
* returns all trust info in group membership details in `trust` field
* adds a `role` model to let groups specify custom roles, with fields `display_name`, `description`, `threshold`, `role_required_for_trust`
* adds notifications and email when somebody got a new role
* works with existing frontend by:
  * preserving `trusted_by` serializer field in group membership info
  * defaulting to 'editor' if not specified

To try it out, use the database shell to create a new `Role` for a given group:
```py
Group.objects.get(id=1).roles.create(
  name='approved', 
  display_name='Bestätigt', 
  description='Wenn man bestätigt ist, kann man tollere Abholungen machen.'
  role_required_for_trust='editor',
)
```


tiltec:
Next, I'll add the bit to actually assign the role when there's at least one trust. And to remove it again when trust is revoked and there's less than one trust left.... I guess that can somehow work...

nicksellen:
> Next, I'll add the bit to actually assign the role when there's at least one trust. And to remove it again when trust is revoked and there's less than one trust left.... I guess that can somehow work...

Great :) Makes sense for me to review it after that.

The notifications might be the more fiddly, I think the ones for getting editor role are very specific to that, but would be useful if they were generalizable, so when roles are customizable per group there is just one set of role/trust notifications.

tiltec:
I wondered if it could make sense to have those notifications somehow user-defined; after all, I have no idea what the role is actually good for. From my perspective, it's just a lower-case string ;)

We probably need those messages:
- human-readable role name
- frontend trust dialog, info about role
- frontend you_got_role notification text (could also be used for backend email)
- frontend role_removed notification text
- ...

This would turn Karrot more into a Content Management System, but could be a useful level of customization. I guess groups would write better messages than we could, because they are closer to the users.

This also resolves the question about translation, groups would provide the messages in their own language.

nicksellen:
> I wondered if it could make sense to have those notifications somehow user-defined; after all, I have no idea what the role is actually good for. From my perspective, it's just a lower-case string ;)

Totally :) That's exactly the kind of direction that makes sense to me too. And more generally having more and more bits of the app that the groups themselves have written. This is part of the same vision of the customizable activity types, place types, etc...

I think there is an intermediate step though, as that's a lot of work to get to that, and it would probably be best to do some design process first to understand the context for the groups better.

So, to my mind a way to notify about people getting _approved_ role is enough for now, whether that's by generalizing the existing _editor_ content, or making new content is up to you :) (or someone else if you aren't doing that bit)

tiltec:
I mean, I wouldn't implement any API method to change the text yet, but still have it in the database so we can adapt to the Luxembourg group needs without re-deploying (but just change the database entry).

Basically, my idea is to set up the code so the Luxembourg group can try out their "approved" role with text of their own choosing (set through us), and if other groups would like to join the beta test, they can do so with fairly low effort for us (just changing a few database fields).

I wouldn't touch the existing messages for editor yet, would leave that for later.

At some point, we can add the API to add roles and change messages. Then it makes sense to adjust the editor messages, as roles would become more prevalent in the groups.

karrot/groups/templates/user_got_role.subject.jinja2
```diff
@@ -0,0 +1 @@
+{% trans group_name=group_name|safe %}You received a new role in {{ group_name }}!{% endtrans %}
```
nicksellen, line 1: Comment only: most email subjects start with `[{{ site_name }}]` - although not the other trust ones.

I think `[{{ group_name }}]` would actually be better, so the identity is more focused on the group, not on the site, might be possible to do it at a higher level, to always add that to the subject, where there is a group set ...

karrot/groups/tests/test_serializers.py
```diff
@@ -16,14 +16,15 @@ def setUp(self):
 
     def test_detail(self):
         serializer = GroupDetailSerializer(self.group, context={'request': MockRequest(user=self.members[0])})
-        self.assertEqual(len(serializer.data.keys()), 23)
+        self.assertEqual(len(serializer.data.keys()), 24)
```
nicksellen, line 19: The most important test of all!

karrot/groups/serializers.py
```diff
@@ -280,6 +318,41 @@ def update(self, instance, validated_data):
...
+    class Meta:
+        model = GroupMembership
+        fields = ('role', )
+
+    # default so it works with non-trust-for-role aware frontend
+    role = serializers.CharField(default=GROUP_EDITOR)
+
+    def validate_role(self, role):
+        group = self.instance.group
+        if role not in group.all_role_names:
```
nicksellen, line 332: I guess this is where we can check if the person giving the trust has sufficient role to give that trust (i.e. only editor (or editor and approved) should be able to give trust for approved role - as per discussion in frontend PR). _and/or_ configurable in the `Role` model....

nicksellen (commented):
Great, I think it's amazing to progress this stuff.

It does feel an intermediate state now, not on your part, but because we're bolting concepts on to our existing system, which I think would benefit from some rationalization, which I think we are capable to manage now! So great to work with what we have right now.

I can't actually approve this, because GitHub tells me it's my PR, because I did the initial work on it :)

nicksellen:
> I wouldn't touch the existing messages for editor yet, would leave that for later.

yup, makes sense!

tiltec (approved):

PR #1171, opened 2021-07-19
Move channel subscription sending into task

tiltec:
Accepting applications is amongst the slowest frequent operations on Karrot, taking multiple seconds (mean 6s, 95perc 20s). Generally, would be nice to speed up write operations somehow :)

I figured it would be good to move more secondary work out of the request, starting with websocket updates. 

I would add a general `notify_channel_subscribers(topic, subscription_filter, make_payload)` task that can be reused for all current websocket notification purposes. It would be used like this:
```python
    notify_channel_subscribers(
        topic='places:place',
        subscription_filter=Q(user__in=place.group.members.all()),
        make_payload=lambda subscription: PlaceSerializer(place, context={'request': MockRequest(user=subscription.user)}).data,
    )
```

The lambda seems a bit clunky, maybe it's better to have a named function there?

We could also have a helper function that takes a serializer and the instance:
```python
notify_channel_subscribers(
    topic='places:place',
    subscription_filter=Q(user__in=place.group.members.all()),
    make_payload=payload_from_serializer(PlaceSerializer, place)
 )

def payload_from_serializer(serializer_class, instance):
  def inner(subscription):
    return serializer_class(instance, context={'request': MockRequest(user=subscription.user)}).data
  return inner
```

Any thoughts about this?


nicksellen:
At some point I experimented with an annotation to do the whole receiver logic in a task. I think quite a lot could be done like that (also maybe makes them less transactional sensitive, and/or can automatically submit the task on transaction complete...).

If that's sounds interesting I might be able to dig up the code again.

I agree lambdas are a bit out of place in python. Plain functions are good though...

tiltec:
> If that's sounds interesting I might be able to dig up the code again.

Yup, sounds interesting! Could be the simpler change.

nicksellen:
I found the branch locally, but I went a bit too far with the abstraction https://github.com/yunity/karrot-backend/commit/bb6bdd32d7425a6afbc912a478f8282cae4b13c6 ... maybe there is something useful in there though. It's not just quite an annotation that runs a receiver in a task (but I think that would be technically possible, they're quite flexible...).

I think what you've done also would work very nicely, so you could have a look/think and decide what makes sense right now..

tiltec:
I'm leaning towards having less abstraction, so I might go with my original approach plus `@on_transaction_commit`, to prevent race conditions (otherwise the task might sometimes run earlier than transaction commit).

But it would be even easier if I could just apply the `db_task` decorator to each signal handler, like this:
```py
@receiver(post_save, sender=Activity)
@on_transaction_commit
@db_task
def activity_date_saved(sender, instance, **kwargs):
   ...
```

Not sure if it can actually work.

Also, we usually have the tasks in a `tasks.py` file and receivers in `receivers.py` - but I don't fully understand the import semantics. `receivers.py` is explicitly imported in `__init__.py`, whereas `tasks.py` is auto-imported by huey.

tiltec:
Tried it, this works in tests:

```py
@receiver(post_save, sender=Feedback)
@on_transaction_commit
@db_task()
def feedback_saved(sender, instance, created, **kwargs):
   ....
```

In tests, I need to slap a lot of context managers on the action under test:

```py
with execute_scheduled_tasks_immediately(), self.captureOnCommitCallbacks(execute=True):
  Activity.objects.process_finished_activities()
```

I wondered whether I should move this to a TestCase mixin, to have it like this:

```py
class FinishedActivityReceiverTest(WSTestCase, ExecuteDeferredTasksMixin):
  ...
```

It would run scheduled tasks and on_commit callbacks for all test code, saving some boilerplate.

nicksellen:
Seems good! I wonder if an abstraction like this could make sense (assuming it's actually technically possible to implement):

```py
@receiver_task(post_save, sender=Feedback)
def feedback_saved(sender, instance, created, **kwargs):
```

Which would wrap @receiver, @on_transaction_commit, and @db_task all in one (or @receiver_transaction_task to make the whole chain clearer?).

I'm fine either way though.

Likewise the context managers could be presumably be combined, and then have:
```py
with deferred_tasks():
  Activity.objects.process_finished_activities()
```

At the moment I think we use it sometimes for only part of the code in a test method right? I wonder if it works if it's over the whole method/class (as it would be for a mixin).

Again, happy with any of the options described...

tiltec:
Yup, both should work! I pushed some experimental code, tests are still failing, most likely for good reasons.

One problem I just saw: `pre_delete` signal handlers sometimes have `None` in the instance id. Seems we just keep a reference to the instance in the `on_transaction_commit` wrapper (as part of kwargs), and then the instance changes when the object is actually deleted from the database.

I guess `on_transaction_commit` should actually do a deep copy of `args` and `kwargs` to handle changing objects?

https://github.com/yunity/karrot-backend/blob/82435082befbd8dedce00bda2738ae5edab03b9a/karrot/utils/misc.py#L7-L11

tiltec:
> I guess on_transaction_commit should actually do a deep copy of args and kwargs to handle changing objects?

This didn't work out because it couldn't copy some thread lock and weakref thing, so I made a smaller change: For `pre_delete` signals, shallow-copy the `instance` and forward it as another signal.

karrot/groups/models.py
```diff
@@ -163,10 +162,7 @@ def accept_invite(self, user, invited_by, invited_at):
         )
 
     def refresh_active_status(self):
-        self.last_active_at = tz.now()
-        if self.status == GroupStatus.INACTIVE.value:
-            self.status = GroupStatus.ACTIVE.value
-        self.save()
+        type(self).objects.filter(id=self.id).update(last_active_at=tz.now(), status=GroupStatus.ACTIVE.value)
```
tiltec, line 165: This avoids triggering `post_save` handlers, to not send websocket updates...

karrot/groups/receivers.py
```diff
@@ -80,6 +81,8 @@ def trust_given(sender, instance, created, **kwargs):
 
         prepare_user_became_editor_email(user=membership.user, group=membership.group).send()
 
+        roles_changed.send(sender=GroupMembership.__class__, instance=membership)
```
tiltec, line 84: Custom signal intended to replace usage of `DirtyFieldsMixin`

karrot/groups/serializers.py
```diff
@@ -426,6 +427,8 @@ def update(self, instance, validated_data):
         notification_type = validated_data['notification_type']
         instance.add_notification_types([notification_type])
         instance.save()
+
+        notification_type_changed.send(sender=GroupMembership.__class__, instance=instance)
```
tiltec, line 431: Custom signal intended to replace usage of `DirtyFieldsMixin`

karrot/subscriptions/receivers.py
```diff
@@ -39,14 +41,20 @@
...
 from karrot.users.serializers import UserSerializer
-from karrot.utils.misc import on_transaction_commit
 
+pre_delete_with_cloned_instance = Signal()
 
-@receiver(post_save, sender=ConversationMessage)
-@on_transaction_commit
+
+@receiver(pre_delete)
+def clone_instance_and_forward(signal, sender, instance, **kwargs):
```
tiltec, line 52: This provides `id` fields to signal receivers that used `pre_delete` before - during object deletion, Django sets the `id` field to `None`, making it unavailable to `pre_delete` receivers that run `on_commit`.

tiltec (commented):
Further changes I did:
- replace `DirtyFieldsMixin` with more custom signals, because it ooesn't work with `on_commit`. We used `DirtyFieldsMixin` to reduce the number of unnecessary websocket messages, but it never seemed like a good solution
- rewrite some tests to aid debugging

karrot/subscriptions/receivers.py
tiltec: Seems this throws some errors in other parts of the code :( Well, maybe I need another solution...

PR #1198, opened 2021-11-20
include calendar name in ics export

amengsk:
ugly hack.
fixes https://github.com/yunity/karrot-frontend/issues/2452
lemme know if you see a better way!


nicksellen (approved):
It's definitely an improvement :+1: 

I wonder if it made sense to have the name as "<group>: <place>" when a place is specified? But fine either way.

I'm assuming clients that subscribe will update the name on the next fetch (once deployed..).

amengsk:
you mean "group: place" instead of "place"?

nicksellen:
> you mean "group: place" instead of "place"?

Exactly, but just an idea. I think either is an improvement.

nicksellen:
I merged already, forgetting there was still an open question! But it's already an improvement, but feel free to make another PR if you think the other way is better ("group: place")

PR #1200, opened 2021-11-26
exposed disabled activities as cancelled in ics export

amengsk:
closes yunity/karrot-frontend#2454


nicksellen (approved):
Look great, thanks!

PR #1203, opened 2021-12-16
Password Policy for New Users

rheidem:
Frontend PR: [karrot-dev/karrot-frontend#2436](https://github.com/karrot-dev/karrot-frontend/issues/2436)

Hello! Submitting this PR to get the backend code for the password policy out there! Please let me know what needs to be fixed and how the continuation of this PR (the frontend) should be completed in tandem with this PR. Thank you

Password Validation Specifics:
- Minimum length of 8 characters
- Must use at least one number
- Must use at least one lowercase letter
- Must use at least one uppercase letter
- Must use at least one special character
- Cannot contain any form of the string "karrot"
- Cannot match a blacklisted password

Note: The "blacklisted" passwords were obtained by [here](https://www.ncsc.gov.uk/static-assets/documents/PwnedPasswordsTop100k.txt) and filtered by the previous criteria to only match against blacklisted passwords that also have a minimum length of 8 characters, contain at least one number, etc.

Other Concerns:
- To ensure the correctness of previous usage of passwords I had to change the default password for seed data from `123` to `K@rr0t1!`. Wasn't sure what would be best, but that password fits our criteria

### DIFF
--- a/karrot/locale/en/LC_MESSAGES/django.po
+++ b/karrot/locale/en/LC_MESSAGES/django.po
@@ -735,9 +735,9 @@ msgstr ""
 msgid "You are already subscribed to this place"
 msgstr ""
 
-#: karrot/userauth/api.py:98 karrot/userauth/serializers.py:94
-#: karrot/userauth/serializers.py:143 karrot/userauth/serializers.py:164
-#: karrot/userauth/serializers.py:194 karrot/userauth/serializers.py:216
+#: karrot/userauth/api.py:98 karrot/userauth/serializers.py:104
+#: karrot/userauth/serializers.py:153 karrot/userauth/serializers.py:174
+#: karrot/userauth/serializers.py:204 karrot/userauth/serializers.py:226
 msgid "We could not send you an e-mail."
 msgstr ""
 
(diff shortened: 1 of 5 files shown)


nicksellen:
Hey! Thanks for your PR :)

From the original issue (https://github.com/karrot-dev/karrot-frontend/issues/2436) there is a link to a django docs to show how to use the password validators that are already included in django --> https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#module-django.contrib.auth.password_validation

It would seem sensible to use those, rather than write custom validators.

We also need to consider translations, and I _assume_ the error messages from the django ones will be already translated.

Also, I think having all these types of validators is no longer best practise for password security (e.g. [thoughtworks/NIST](https://www.thoughtworks.com/en-gb/radar/techniques/naive-password-complexity-requirements)) (although probably a lot of guides still recommend them).

This one https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#tip5-password-collection concludes:

> Complexity requirements provide no defence against common attacks and should not be used.

they do say:

> you should specify a minimum password length, to prevent very short passwords from being used

So, from my quick scan of that page I would go with:
- `UserAttributeSimilarityValidator`
- `MinimumLengthValidator` (with length 8, the default)
- `CommonPasswordValidator`

We'll also need to consider how it appears in the frontend, if the errors are clear and/or translated.

nicksellen:
Hey! Thanks for your PR :)

From the original issue (https://github.com/karrot-dev/karrot-frontend/issues/2436) there is a link to a django docs to show how to use the password validators that are already included in django --> https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#module-django.contrib.auth.password_validation

It would seem sensible to use those, rather than write custom validators.

We also need to consider translations, and I _assume_ the error messages from the django ones will be already translated.

Also, I think having all these types of validators is no longer best practise for password security (e.g. [thoughtworks/NIST](https://www.thoughtworks.com/en-gb/radar/techniques/naive-password-complexity-requirements)) (although probably a lot of guides still recommend them).

This one https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#tip5-password-collection concludes:

> Complexity requirements provide no defence against common attacks and should not be used.

they do say:

> you should specify a minimum password length, to prevent very short passwords from being used

So, from my quick scan of that page I would go with:
- `UserAttributeSimilarityValidator`
- `MinimumLengthValidator` (with length 8, the default)
- `CommonPasswordValidator`

We'll also need to consider how it appears in the frontend, if the errors are clear and/or translated.

How does that sound?

PR #1207, opened 2021-12-31
Add @mentions

nicksellen:
Closes #2407 
Forum topic: https://community.karrot.world/t/mentions-on-karrot/732/8
Frontend https://github.com/karrot-dev/karrot-frontend/pull/2479

This adds two things:
1. add username field to user (so there is something to mention) (case insensitive unique)
2. notifies users by email and bell when mentioned by username in non-private conversations

It does not (but could/should be added in the future):
- allow changing username (should also rename mentions in messages)
- allow signing in by username (still only by email)

It automatically creates usernames for existing users, prioritising recently seen users:
- split display name, take first word, lowercase it, remove non-ascii chars
- display name with spaces removed, lowercased, remove non-asii chars
- part of the email address before the @ sign, lowercased, remove non-ascii chars
- random username

New users have to sign up with a unique username of their choosing.


tiltec (commented):
Great step forward, thank you for working on this!

Some things that came to my mind while reading:

Seems a bit limiting to have user names coded directly into the message - changing user names might be quite complicated if we ever want to have this feature.

How about a way to disable mention notification emails? I guess this opens up the "better notification settings" discussion again - should the setting be per group or site-wide, etc...

What about app push notifications? Maybe not getting used much, but it might not be much work to make it complete.

When a mention gets removed, should we remove the bell notification as well? (Email is harder to undo...)

Should the user name become a link to the profile also in emails? I didn't look at the frontend yet, but I assume this is what happens there.

nicksellen:
Thanks for the review, super helpful :)

> Seems a bit limiting to have user names coded directly into the message - changing user names might be quite complicated if we ever want to have this feature.

I explored a few options [over here](https://github.com/karrot-dev/karrot-frontend/issues/2407#issuecomment-969236821), basically:
- we only have a plain text editor, so user will see whatever the markup is
- `@id` (e.g. `@4243`) would mean nothing to the user
- any other markup is even more complicated than usernames

Changing usernames will be OK! We can rewrite the message content (discourse does it). We have table in the db saying which messages have users mentioned, so know exactly what needs replacing. Can also include descriptions, or not (but they are not tracked in db).

> How about a way to disable mention notification emails? I guess this opens up the "better notification settings" discussion again - should the setting be per group or site-wide, etc...

Yes, it does open that topic, which I didn't want to. And it seems a reasonable starting point to always notify on mention... (better than missing them). I think mentioning really gives a sense to the author they will be notified.

> What about app push notifications? Maybe not getting used much, but it might not be much work to make it complete.

Yup, good idea. I'll add them.

> When a mention gets removed, should we remove the bell notification as well? (Email is harder to undo...)

Good point, yes, bell can go...

> Should the user name become a link to the profile also in emails? I didn't look at the frontend yet, but I assume this is what happens there.

It could, although there is already a link to the user with the message author text. Would be a small change in the backend markdown mentions addon thing.

PR #1210, opened 2022-01-09
[CI] update dependencies

tiltec:
Good morning, I found some fresh dependencies for you!

This Pull Request contains updates done by `pip-compile --upgrade --no-annotate` and
in `mjml`, `yarn upgrade --latest`.

Please check the changes before merging and consider trying out the branch locally to verify if it works.


nicksellen:
Oh, this tries to upgrade to django 4, which needs python 3.8, we have 3.7 in CI and in prod...

PR #1214, opened 2022-02-14
Increase activity page size

nicksellen:
This does two things:
- increases activity page size from 800 to 1200 (temporary fix, until we address https://github.com/karrot-dev/karrot-frontend/issues/2388)
- adds an index for activities date, I'm not 100% sure it'll use it as it's on a range type, and had to use a GIST one... and well, just more complex than usual case


nicksellen:
Requested review, but I'll merge by tomorrow in any case (maybe before), as I'm going away for the week.

tiltec (approved):

nicksellen:
Well, it uses the new date gist index, at least for a basic query with date filter:

```sql
explain analyze select * from activities_activity where date && pg_catalog.tstzrange('2022-02-14T13:14:37.093000+00:00'::timestamptz, NULL, '[)');
```

```
 Bitmap Heap Scan on activities_activity  (cost=55.47..2817.56 rows=1444 width=128) (actual time=4.676..12.180 rows=5904 loops=1)
   Recheck Cond: (date && '["2022-02-14 14:14:37.093+01",)'::tstzrange)
   Heap Blocks: exact=1461
   ->  Bitmap Index Scan on activities__date_5ffa2e_gist  (cost=0.00..55.11 rows=1444 width=0) (actual time=4.106..4.106 rows=5904 loops=1)
         Index Cond: (date && '["2022-02-14 14:14:37.093+01",)'::tstzrange)
 Planning Time: 0.279 ms
 Execution Time: 12.996 ms
(7 rows)
```

PR #1229, opened 2022-06-26
Add token auth API for ICS export

tiltec:
Allows ICS export without username/password, by using a token query parameter.

- `POST /api/activities/ics_token_refresh/` returns a new token; invalidates old tokens, there can only be one per user
- `GET /api/activities/ics_token/` returns the current token
- `GET http://localhost:8000/api/activities/ics/?token=N77xPKKcx2fpv240YvqvxufbmQNJJ2rpJyugGRj3` does the ICS export; filter query params are also possible

Related to https://github.com/karrot-dev/karrot-frontend/issues/2541


karrot/activities/api.py
```diff
@@ -281,11 +294,29 @@ def ics_detail(self, request, pk=None):
         renderer_classes=(ICSCalendarRenderer, ),
         serializer_class=ActivityICSSerializer,
         url_path='ics',
-        authentication_classes=[BasicAuthentication, SessionAuthentication],
+        authentication_classes=[BasicAuthentication, SessionAuthentication, ICSQueryTokenAuthentication],
```
tiltec, line 297: I think this means it tries the other auth mechanisms before reading the token.

nicksellen: It does. I just tried it, sending basic auth, and it uses that first, then the token.

Great it still works with basic auth :)

nicksellen (approved):
Great, looks good!

amengsk (approved):
hey thats really nice!
thanks a bunch for implementing this @tiltec

PR #1235, opened 2022-07-28
Add more activity filters and pagination options

nicksellen:
As part of the work for frontend activity pagination / filtering, using vue-query, the backend needs some updates. This should be backwards compatible with current frontend as it only adds _more_ options, and does not take any away.

For activities, you can now:
- pass in a `page_size` param to say how many results per page (default is 1200, as current frontend has no pagination)
- pass in values the filters that we used to do in front-end only, so:
  - `activity_type`
  - `slots` (including a new "joined" option, that does same as "joined" filter param...)

This would be useful to review/merge as soon as possible, as until then I can't see how well the new frontend is doing in realworld examples...

### DIFF
--- a/karrot/activities/api.py
+++ b/karrot/activities/api.py
@@ -203,6 +203,7 @@ class ActivityPagination(CursorPagination):
     Could be reduced and add pagination handling in frontend when speed becomes an issue"""
     # TODO: we *really* need to add that pagination to the frontend! https://github.com/karrot-dev/karrot-frontend/issues/2388
     page_size = 1200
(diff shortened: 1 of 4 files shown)


nicksellen:
Pinging both @amengsk and @tiltec for review, happy with anyone. You'll notice I didn't write any tests, but would be happy to told off about that, and then I would go away and write some :)

tiltec (approved):

tiltec:
Code looks good, test failure seems due to flakiness?

Would be good if you add a test for each slot filter value!

PR #1239, opened 2022-08-16
Add more bootstrap data and other tweaks

nicksellen:
This adds more stuff into the boostrap endpoint, with a new "fields" parameter so the frontend can choose which components to receive. This is to allow it to evolve over time, rather than always get everything. When omitting the parameter it'll default to what it did before there was a parameter option.

Other tweaks:
- allow to get activities individually (i.e. not-list actions) from inactive places, this was needed to fetch the activity for activity chats where the place is now inactive...
- only send activity update websocket events after transaction complete (otherwise refetches triggered from the updates will have previous data)
- enable x-forwarded-for headers in local dev (so if using https during dev, we can get https:// proto in image urls)
- add "push notifications are enabled!" push message, which helps to see if it actually worked...
- allow filtering issues by multiple statuses at once (useful for getting "not-ongoing" ones...

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -389,6 +389,11 @@ HOSTNAME = options['SITE_URL']
 SITE_NAME = options['SITE_NAME']
 MEDIA_ROOT = options['FILE_UPLOAD_DIR']
 
+if is_dev:
+    # in prod daphne (and I guess uvicorn) handle this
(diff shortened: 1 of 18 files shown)


tiltec (approved):
Looks good!

PR #1241, opened 2022-09-19
Add public activities

nicksellen:
Closes https://github.com/karrot-dev/karrot-frontend/issues/2583

Frontend PR: https://github.com/karrot-dev/karrot-frontend/pull/2584

The idea is:
- add `is_public` and `public_id` fields to activity
- `public_id` is a UUID
- add a banner image field to activity, mostly intended to use for public activities, but doesn't need to be limited to that...
- add `PublicActivity*` API/serializers/etc. available at `/api/public-activities/*`
- the idea to keep it separate is that it's a much more restricted set of fields available...
- it includes nested place/group/activityType info, so the page can render from one request .. and updating live is (probably) less important


karrot/activities/serializers.py
```diff
@@ -390,10 +436,19 @@ def update(self, activity, validated_data):
                     ParticipantType.objects.create(activity=activity, **entry)
 
         update_data = validated_data.copy()
+
+        if 'is_public' in update_data:
+            if update_data['is_public']:
+                if not activity.public_id:
```
tiltec, line 442:
```suggestion
        if 'is_public' in update_data and update_data['is_public'] and not activity.public_id:
```

might be nicer with less nesting :)

tiltec (approved):
Looks good!

karrot/activities/serializers.py
nicksellen: Ah yeah :) An earlier iteration had some `else` clauses... (before I would set `public_id` to `null` when `is_public` was false... but thought it might be nice to keep the id, so you can enable/disable public without the links breaking...).

config/settings.py
```diff
@@ -328,7 +328,11 @@
         ('full_size', 'url'),
         ('200', 'thumbnail__200x200'),
         ('600', 'thumbnail__600x600'),
-    ]
+    ],
+    'activity_banner_image': [
+        # TODO: work out what to do here...
+        ('full_size', 'url'),
```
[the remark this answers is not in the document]
nicksellen: Yes, I agree. 800x300 is actually the size i decided on for the banner, so could have put it in...  will leave it for now.

PR #1254, opened 2023-04-23
[CI] update dependencies

tiltec:
Django 4 and shiny new stuff!

I put off some of the actual upgrading work, such as upgrading from pytz to zoneinfo (`USE_DEPRECATED_PYTZ`) and ignoring future deprecation warnings.

We might need to change how we set CSRF_TRUSTED_ORIGINS, it now wants a https in front: https://docs.djangoproject.com/en/4.2/releases/4.0/#csrf-trusted-origins-changes


nicksellen:
Very nice to have this upgrade work done! :partying_face: :star: 

For the CSRF config changes, it says:

> A system check detects any required changes.

So at least that'll prompt us... we will need to change them in our yuca config, as they are like `'.karrot.world' at the moment.

As we have our _options_ layer that maps between env variables and stuff in settings.py would could adapt them, but maybe easy enough to just let it break and fix things.

I notice the tests failed with:

> No time zone found with key Europe/Kyiv

Random internet person here (https://stackoverflow.com/a/73333278) fixed it with:

> Downgrading the pytz version from 2022.2 to 2022.1 seems to have solved this issue for me.

.. but not sure if that is relevant for us.

Hmmm. For another timezone issue I did hack it on the server to fix the timezone files, but seems maybe fixable otherwise.

nicksellen:
Ah, I wonder if the automatic update broke your nice upgrades?

nicksellen (approved):
Woo! :partying_face: .. very nice to have this.

This review has the wrong code now because of the CI auto-push, but https://github.com/karrot-dev/karrot-backend/pull/1258 is the one with the changes, which I just merged.

Thanks!

PR #1255, opened 2023-04-24
Support message attachments

nicksellen:
Early WIP

Currently messages can have _images_ associated with them, this extends that so that messages can have _attachments_ of any kind, including images.


nicksellen:
Closed in favour of https://github.com/karrot-dev/karrot-backend/pull/1256

PR #1264, opened 2023-07-25
Remove from all conversations when leaving group

nicksellen:
Fixes https://github.com/karrot-dev/karrot-frontend/issues/2571

There are multiple types of chat where users are not always removed from the chat when leaving the group. Activity chats is one case, which was an issue for past activity chats.

### DIFF
--- a/karrot/places/receivers.py
+++ b/karrot/places/receivers.py
@@ -1,4 +1,3 @@
-from django.contrib.contenttypes.models import ContentType
 from django.db.models.signals import post_save, pre_delete
 from django.dispatch import receiver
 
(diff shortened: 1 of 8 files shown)


karrot/places/receivers.py
```diff
@@ -47,6 +47,8 @@ def subscription_removed(sender, instance, **kwargs):
 @receiver(pre_delete, sender=GroupMembership)
 def group_member_removed(sender, instance, **kwargs):
     membership = instance
+    membership.group
```
tiltec, line 50: I guess this avoids lazy access in the loop?

tiltec (approved):
Seems good! I guess we can make a database backup right before deploying to production, just to be safe :) Also, messages are not lost when a participant is deleted, so it's mostly reversible.

karrot/places/receivers.py
nicksellen: Ah, that was unintentional. I think maybe one of the `pre-commit` hooks removed itself assignment to a local variable, but left it there (incase it had side effects?). Removed in latest commit.

PR #1270, opened 2023-09-07
Do not create participant when notification settings are not provided

tiltec:
Fixes https://github.com/karrot-dev/karrot-frontend/issues/2675


nicksellen (approved):
Looks good, let's see if we get any of those errors in sentry.

PR #1279, opened 2023-12-18
Switch to ruff linter/formatter

nicksellen:
It's soooo fast! And combines many tools that were previously separate.

https://docs.astral.sh/ruff/

I reformatted and fixed everything using it... so a massive diff.


nicksellen:
Closed in favour of https://github.com/karrot-dev/karrot-backend/pull/1282

PR #1290, opened 2024-03-06
Support python 3.11

nicksellen:
Was limited by cchardet dependency before

I switched the talon dependency from @tiltec's fork to the zulip fork, as in addition to not requiring the machine learning deps, it also fixes the cchardet dep issue.


nicksellen:
I'll merge this now, although we'll still be building the .pyz files at https://download.karrot.world/ with python 3.9 for now.

Actually deploying to 3.11 will be a future step...

